PR #5, opened 2020-07-30
[Task #1345] Fix lint warnings.

ledsoft:
https://kbss.felk.cvut.cz/redmine/issues/1345

### DIFF
--- a/src/component/changetracking/AssetHistory.tsx
+++ b/src/component/changetracking/AssetHistory.tsx
@@ -6,7 +6,7 @@ import {Table} from "reactstrap";
 import {connect} from "react-redux";
 import {injectIntl} from "react-intl";
 import {ThunkDispatch} from "../../util/Types";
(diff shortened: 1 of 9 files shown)


psiotwo (commented):
Building it on my machine still ends up with these warnings:

./src/component/public/vocabulary/VocabularyManagement.tsx
  Line 20:8:  React Hook React.useEffect has a missing dependency: 'props'. Either include it or remove the dependency array. However, 'props' will change when *any* prop changes, so the preferred fix is to destructure the 'props' object outside of the useEffect call and refer to those specific props inside React.useEffect  react-hooks/exhaustive-deps

./src/model/changetracking/PersistRecord.ts
  Line 8:5:  Useless constructor  @typescript-eslint/no-useless-constructor

ledsoft:
Should be fixed.

psiotwo (approved):
lgtm

PR #6, opened 2020-07-30
[R1065] initial implementation of skos:altLabel and skos:prefLabel on term detail

psiotwo:
https://kbss.felk.cvut.cz/redmine/issues/1065

Included:
- altLabels as a part of the header of term detail screen
- altLabels + hiddenLabels editable in term detail edit
- hiddenLabels not part of the term detail view

Not Included:
- fulltext search results

### DIFF
--- a/src/component/term/TermDetail.tsx
+++ b/src/component/term/TermDetail.tsx
@@ -93,7 +93,7 @@ export class TermDetail extends EditableComponent<TermDetailProps> {
         const buttons = this.getButtons();
         return <div id="term-detail">
             <HeaderWithActions title={
-                <>{term.label}<CopyIriIcon url={term.iri as string}/></>
+                <>{term.label}<CopyIriIcon url={term.iri as string}/><br/><h6>{Utils.sanitizeArray(term.altLabels || []).join(", ")}</h6></>
             } actions={buttons}/>
 
             {this.state.edit ?
--- a/src/component/term/TermDetail.tsx
+++ b/src/component/term/TermDetail.tsx
@@ -93,7 +93,7 @@ export class TermDetail extends EditableComponent<TermDetailProps> {
         const buttons = this.getButtons();
         return <div id="term-detail">
             <HeaderWithActions title={
-                <>{term.label}<CopyIriIcon url={term.iri as string}/><br/><h6>{Utils.sanitizeArray(term.altLabels || []).join(", ")}</h6></>
+                <>{term.label}<CopyIriIcon url={term.iri as string}/><br/><h6>{Utils.sanitizeArray(term.altLabels).join(", ")}</h6></>
             } actions={buttons}/>
 
             {this.state.edit ?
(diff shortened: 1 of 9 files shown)


src/component/term/TermDetail.tsx
```diff
@@ -93,7 +93,7 @@ export class TermDetail extends EditableComponent<TermDetailProps> {
         const buttons = this.getButtons();
         return <div id="term-detail">
             <HeaderWithActions title={
-                <>{term.label}<CopyIriIcon url={term.iri as string}/></>
+                <>{term.label}<CopyIriIcon url={term.iri as string}/><br/><h6>{Utils.sanitizeArray(term.altLabels || []).join(", ")}</h6></>
```
ledsoft: Calling `Utils.sanitizeArray(term.altLabels || [])` is superfluous, sanitization ensures that an array is returned, so there's no need for the conditional argument.

src/component/term/TermMetadataEdit.tsx
```diff
@@ -99,6 +108,20 @@ export class TermMetadataEdit extends React.Component<TermMetadataEditProps, Ter
                                          help={i18n("term.label.help")}/>
                         </Col>
                     </Row>
+                    <Row>
+                        <Col xs={12}>
+                            <StringListEdit list={this.state.altLabels || []}
```
ledsoft: Why is the array sanitized here, when it is also sanitized in the `StringListEdit` component? This way, the list is IMHO sanitized unnecessarily multiple times. I suggest leaving the sanitization in the `StringListEdit` component, and not doing it here. Also, we should probably stick to a single method of sanitization - either the `value || []` or `Utils.sanitizeArray` (the second version being able to handle accidental singular value as well).

src/component/term/TermMetadataEdit.tsx
```diff
@@ -99,6 +108,20 @@ export class TermMetadataEdit extends React.Component<TermMetadataEditProps, Ter
...
+                                            onChange={this.onAltLabelsChange}
+                                            i18nPrefix={"term.metadata.altLabels"}/>
+                        </Col>
+                    </Row>
+                    <Row>
+                        <Col xs={12}>
+                            <StringListEdit list={this.state.hiddenLabels || []}
```
ledsoft: DTTO

src/component/term/TermDetail.tsx
psiotwo: fixed

src/component/term/TermMetadataEdit.tsx
psiotwo: fixed

src/component/term/TermMetadataEdit.tsx
psiotwo: fixed

ledsoft (approved):
LGTM

PR #7, opened 2020-07-31
[R1065] fixes to layout for high resolution

psiotwo:


src/component/misc/StringListEdit.scss
```diff
@@ -1,3 +1,7 @@
-.no-shadow-box {
-  box-shadow: none !important;
+.no-bottom-margin {
+  margin-bottom: 0rem !important;
+}
+
+.form-group-bottom-margin {
```
ledsoft, line 5: Doesn't regular form-group cover the bottom margin?

ledsoft (approved):
See comments.

src/component/misc/StringListEdit.scss
psiotwo: it does, but there is a problem with the textual hint when being part of the group (on large resolutions the hint is layouted on the same row as the input/button group).

PR #9, opened 2020-08-06
[Bug #1351] Save original navigation target when redirecting to login and go to it after login

ledsoft:
https://kbss.felk.cvut.cz/redmine/issues/1351


psiotwo (approved):
lgtm

PR #10, opened 2020-08-07
[R1348] adding alt/hidden label fields to term creation

psiotwo:


blcham (approved):
seems ok, but not sure if it will work in annotator (let's see after merge as it was not explicitly specified in the ticket).

PR #11, opened 2020-08-10
[Bug #1353] Fix regression which prevented users from updating their password

ledsoft:
https://kbss.felk.cvut.cz/redmine/issues/1353


blcham (approved):
seems ok

PR #12, opened 2020-08-10
[Bug #1359] Reset annotation term to prevent stale term being displayed.

ledsoft:
https://kbss.felk.cvut.cz/redmine/issues/1359


blcham (approved):
seems ok

PR #13, opened 2020-08-11
[Bug #1358]

ledsoft:
Replace annotation node with text node in list of DOM nodes when it is a top level node on annotation removal.

https://kbss.felk.cvut.cz/redmine/issues/1358


blcham (approved):
seems ok

PR #15, opened 2020-09-03
[Bug #1374] Use deploymentName argument in Webpack build

ledsoft:
https://kbss.felk.cvut.cz/redmine/issues/1374


psiotwo (approved):
lgtm

PR #19, opened 2020-09-22
[R1360] annotator refresh

psiotwo:


blcham (dismissed):
lgtm, although it is hard to judge. Note that some checks in build failed.

PR #20, opened 2020-10-08
1400 multilingual atts ui

ledsoft:
https://kbss.felk.cvut.cz/redmine/issues/1400

Frontend for displaying multilingual attributes + changes required in the model (will apply also to editing of multilingual attributes).


psiotwo (approved):
lgtm

PR #28, opened 2020-10-21
[R1418] removing term occurrences (incorrect)

psiotwo:


blcham (approved):
Works as expected. Approving.

There are however some issue with the semantics of "navrzeny vyskyt" that we can discuss on the meeting (If I approve all "suggested occurrences", their count does not change.)

PR #29, opened 2020-10-23
1347 multilingual attribute editing

ledsoft:
https://kbss.felk.cvut.cz/redmine/issues/1347


psiotwo (approved):
lgtm, but a few comments:
- I don't know how to delete a language - not very important however
- The button "Remove" is quite confusing now - it might look like, one wants to delete just the translation in the selected language, not the whole concept.

ledsoft:
1. Fair enough, I'll add a remove icon to the tabs
2. If you mean in edit mode, we  can move the button to the bottom of the form to the other action buttons

PR #30, opened 2020-10-27
[R1365] simplifying test results UI

psiotwo:
- removing score from search results
- removing heading of the search screen
- changing font of the count info
- moving clear search button to the search field itself


src/component/search/label/NavbarSearch.tsx
```diff
@@ -108,7 +108,9 @@ export class NavbarSearch extends React.Component<NavbarSearchProps, NavbarSearc
 
         const clearIcon = (
             <InputGroupAddon addonType="append" id="search-reset" onClick={this.resetSearch} className="float-right">
-                <Button title={this.props.i18n("search.reset")} color="outline-danger" style={{zIndex : 5}}><FaTimes/></Button>
```
ledsoft: Why is there a z-index set? I don't see any difference w and w/o it...

psiotwo: Left-border of the button does not render black without it.

ledsoft: I can see the border without any problems even when I remove the style. The addon button itself has z-index 2

psiotwo: This is what I see without it 
image

Chromium Version 86.0.4240.111 (Official Build) snap (64-bit)

ledsoft: Ok, nevermind, I see it now, too (although it could be intentional in reactstrap).

ledsoft (approved):

PR #31, opened 2020-10-29
1347 multilingual attribute editing

ledsoft:
https://kbss.felk.cvut.cz/redmine/issues/1347 - allow removal of translation in term.


psiotwo (commented):
Removing Czech language ends up with 
image


https://deploy-preview-31--termit-dev.netlify.app/#/vocabularies/pokusny-dokumentovy-slovnik/terms/bbedrich?namespace=http://onto.fel.cvut.cz/ontologies/slovnik/&includeImported=false

ledsoft:
Fixed

psiotwo (commented):
Language removal works, but I am not able to test it fully, as saving the updated resource does not work

I am getting "Text analysis invocation failed." without having changes persisted.

ledsoft:
The problem with failed text analysis is apparently due to the conflict between frotend on Netlify, which uses github version with multilingual strings and backend on KBSS, which is from development branch and does not use multilingual strings yet.

ledsoft:
Correction: the problem is that the text analysis service uses language configured in server deployment (cs in our case), but if you don't provide definition in that language, annotace fails, because its input is null. I think annotace should be able to handle empty input for analysis.

psiotwo (approved):
tested, lgtm

psiotwo (approved):

PR #33, opened 2020-10-29
[Bug #1429] Do not display search results preview in public search-related views (e.g., search results).

ledsoft:
https://kbss.felk.cvut.cz/redmine/issues/1429


psiotwo (approved):
lgtm

PR #35, opened 2020-10-30
[Bug #1417] Ensure document constructor initializes all relevant fields.

ledsoft:
Fixes https://kbss.felk.cvut.cz/redmine/issues/1417

### DIFF
--- a/src/model/Document.ts
+++ b/src/model/Document.ts
@@ -26,6 +26,7 @@ export default class Document extends Resource implements DocumentData {
 
     constructor(data: DocumentData) {
         super(data);
+        this.vocabulary = data.vocabulary;
         this.files = Utils.sanitizeArray(data.files).map(fd => new File(fd));
     }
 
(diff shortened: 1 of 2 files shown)


src/model/Document.ts
```diff
@@ -26,6 +26,7 @@ export default class Document extends Resource implements DocumentData {
 
     constructor(data: DocumentData) {
         super(data);
+        this.vocabulary = data.vocabulary;
```
psiotwo, line 29: ltgm.

Just a note - to me, we use lots of questionmarked fields in our API (e.g. the field vocabulary? in this class), which make validation difficult.

psiotwo (approved):
lgtm

src/model/Document.ts
ledsoft: Yeah, well there can be standalone documents which do not have a vocabulary, so I don't see much choice (apart from having a subclass which would contain the vocabulary reference).

PR #36, opened 2020-10-30
[Bug #1349] Do not reset selected term in global store when tree select input is cleared in term detail.

ledsoft:
Fixes https://kbss.felk.cvut.cz/redmine/issues/1349


psiotwo (approved):
lgtm

PR #46, opened 2020-11-10
1441 annotator localized term links

ledsoft:
Resolves https://kbss.felk.cvut.cz/redmine/issues/1441


blcham (approved):
works as expected.

PR #47, opened 2020-11-11
[Bug #1443] Remove term occurrence on term creation cancel only if it…

ledsoft:
… was not suggested or assigned an existing term.

Resolves https://kbss.felk.cvut.cz/redmine/issues/1443


blcham (approved):
Works as expected.

PR #48, opened 2020-11-11
[#1350] Visualize term metadata attributes even with empty values.

Saeedla:


ledsoft (approved):
LGTM. I removed display of definition source from the public view, it is always empty there (security reasons), so there's no point showing the attribute.

PR #49, opened 2020-11-11
[R1430] making preflabel validation when switching languages consistent

psiotwo:


ledsoft (approved):
I'd prefer some tests, but I'm getting used to...

PR #50, opened 2020-11-12
[R1442] make analysis button depend on the default vocabulary provided.

psiotwo:


blcham (approved):
Works. Much more intuitive than the previous state.

PR #52, opened 2020-11-19
1418 disabling term occurrences

psiotwo:


blcham (commented):
Sorry to prolonging the ticket, but you removed more than was expected :(((, see pictures in the ticket (https://kbss.felk.cvut.cz/redmine/attachments/257/Selection_295.jpg). You accidentally removed column "assignment". Otherwise everything works as expected.

Moreover, just please check that this is correct:
￼        "term.metadata.assignments.assignment": "Assigned to the resource",
￼        "term.metadata.assignments.assignment": "Anotován tímto pojmem?",

You have question mark only in czech language.

blcham (approved):
Works as expected.

PR #53, opened 2020-11-19
1430 validation of preflabel

psiotwo:


src/component/term/TermMetadataEdit.tsx
```diff
@@ -58,8 +61,15 @@ export class TermMetadataEdit extends React.Component<TermMetadataEditProps, Ter
     public onPrefLabelChange = (prefLabel : string) => {
         const update = {};
         update[this.props.language] = prefLabel;
-        this.setState({labelExists: false, label: Object.assign({}, this.state.label, update)});
-        if (prefLabel.toLowerCase() === getLocalized(this.props.term.label).toLowerCase()) {
+        const updateLabelExists = {};
+        updateLabelExists[this.props.language] = false;
+        this.setState({
+            label: Object.assign({}, this.state.label, update),
+            labelExist: Object.assign({}, this.state.labelExist, updateLabelExists)
```
ledsoft: Why is the `labelExist` attribute initialized as Map, when it is on the first label edit replaced by a plain object?

ledsoft (commented):
This does not resolve the same issue in the term creation form (yes, we have two components which are to a large extent duplicated).

psiotwo:
> This does not resolve the same issue in the term creation form (yes, we have two components which are to a large extent duplicated).

Thanks for reminder. I tried to refactor to cover all cases.

src/component/term/TermMetadataCreateForm.tsx
```diff
@@ -47,20 +46,13 @@ export class TermMetadataCreateForm extends React.Component<TermMetadataCreateFo
...
-    }
-
     private onLabelChange = (e: React.ChangeEvent<HTMLInputElement>) => {
         this.onPrefLabelChange(e.currentTarget.value);
     };
 
     private onPrefLabelChange = (label : string) => {
         const change = Object.assign({}, this.props.termData.label);
         change[this.props.language] = label;
-        this.props.onChange({label: change});
```
ledsoft, line 63: I'm sorry for keeping this blocked, but:
- Why do you create a `change` object, which is not used anywhere besides the assignment
- And then use the `checkLabelUniqueness` method, which IMHO should be a validation-only function, for the side-effect of updating the parent state via the onChange method?

ledsoft (commented):
See the inline comment.

ledsoft (approved):
LGTM

PR #55, opened 2020-11-23
[#1451 Fix] Check for source shape of the rule only if it exists.

Saeedla:
https://kbss.felk.cvut.cz/redmine/issues/1451


blcham (approved):
LGTM

PR #57, opened 2020-11-24
1449 bug definition source selection

ledsoft:
https://kbss.felk.cvut.cz/redmine/issues/1449


blcham (approved):
lgtm and works correctly.

PR #64, opened 2020-12-10
1464 validation redesign

psiotwo:
- moving validation results to term detail
- adding badge to term tree


blcham:
Combining badges in the term tree looks nice and is useful within our main use-case. Just not sure that I would like to see it also while viewing term (on the right of the view) as well. **What value it should give me?** Not sure now how the validation score is computed but I can imagine that such a badge could help me more in the selection of parent of the term in "term edit view" (if the score would depend on the score of a parent).

For the other use cases, it is now rather confusing until we have a way to configure a set of validation rules (per vocabulary) and their related visualization in the badge.

[1] Also, I feel like there should be multiple attributes/flags within the term tree like this one (e.g. "is terms used within documents?", "is it approved/draft?", "how many unresolved comments are on the term?"). For that, I think we might end up rather having a table (as in resources view, vocabularies view) with flags to the right and filtering based on those flags. Until then I like the solution.

psiotwo:
> Just not sure that I would like to see it also while viewing term (on the right of the view) as well. **What value it should give me?** Not sure now how the validation score is computed but I can imagine that such a badge could help me more in the selection of parent of the term in "term edit view" (if the score would depend on the score of a parent).

I think it should indeed be removed from this view. It has no value there.

blcham (approved):
LGTM

PR #66, opened 2020-12-15
Generate identifier race condition

### LINKED ISSUE #65: Race conditions on term creation
At the moment, TermIt response time (probably due to the underlying SPARQL endpoint) is really slow. But that is not the issue. It in fact uncovered a race condition. I was creating a term "Akciová společnost". So I filled in "Akciová společnost" into the term name field, filled in the term source and term type and clicked on "Vytvořit a začít nový".

Only then I saw the warning that there was already a term named "Akciová společnost" in my vocabulary. But that is not all. The term was created anyway, and its IRI is:
`https://slovník.gov.cz/legislativní/sbírka/304/2013/pojem/akcio`.

When investigating, I found out that the IRI of my original term is also wrong:
`https://slovník.gov.cz/legislativní/sbírka/304/2013/pojem/akciová-společnos`

There seems to be a possibility of a delay between filling the term name and creating the IRI for the term causing the term to be created with an unintended IRI.

ledsoft:
Fixes https://github.com/opendata-mvcr/termit-ui/issues/65. Done in KBSS repository as it affects both versions.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -12164,6 +12164,11 @@
         "language-subtag-registry": "~0.3.2"
       }
     },
+    "last": {
+      "version": "1.1.0",
+      "resolved": "https://registry.npmjs.org/last/-/last-1.1.0.tgz",
+      "integrity": "sha1-Bm6kIoftn5q2jv+qZHQn1ZHyZSU="
+    },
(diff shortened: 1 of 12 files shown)


psiotwo (approved):
lgtm

PR #67, opened 2020-12-16
hints changed to recommended

MichalMed:
closes issue 1466 or 1465 (since KBSS does not run, I do not know which one)
replaces Nepovinné with Doporučené and Optional with Recommended in hints for four attributes influencing term overall score


ledsoft (approved):
LGTM

PR #70, opened 2021-01-06
Feature/remember table page size

ledsoft:
Remember table page size so that user does not have to select it over and over again. Also allow showing all items on one page.


psiotwo (approved):
lgtm

PR #71, opened 2021-01-06
[1465] displaying validation errors in term detail edit

psiotwo:
https://kbss.felk.cvut.cz/redmine/issues/1465


MichalMed (changes requested):
Functionality is OK, design is too violent. Please remake according to the redmine issue #1465

MichalMed:
ok, it could be. It still seems quite aggresive to me, but better than before. Functionality is perfect.

PR #74, opened 2021-01-08
[1381] creating resource redesigned (files only as parts of document)

psiotwo:
https://kbss.felk.cvut.cz/redmine/issues/1381

@ledsoft mainly please check - I removed reloading resources after creating resource. Seems to work for me.


ledsoft (approved):
Looks ok, but:
- File/document removal is now quite unintuitive. You have to know that clicking on a file in a document will take you to its detail (which was IMHO fine before, when Files were first-class resources). After that, you can remove the file. Then you can remove the document.
- When removing a file, I would expect to be forwarded to the document that contained it, not the list of all resources.

Maybe these issues are for another task. And maybe it's just me who's confused.
Also yes, there's no need to reload resources on resource creation since the resource list loads them automatically on mount. I suspect this is a remnant of the tree component use.

psiotwo:
> Looks ok, but:
> 
> * File/document removal is now quite unintuitive. You have to know that clicking on a file in a document will take you to its detail (which was IMHO fine before, when Files were first-class resources). After that, you can remove the file. Then you can remove the document.

I agree. I am just implementing the removal as a part of 1378. Files in all the lists will not be clickable, file detail will not exist. Actions performed on File detail will be performed directly in the list, except of Analyze, which I will propose to remove in favour of the same action in the Annotator. 

The only thing that won't be possible will be file renaming which however I think is not a big deal anyway. If yes, we can do some inplace renaming in the File list directly.

> * When removing a file, I would expect to be forwarded to the document that contained it, not the list of all resources.
I think this will be resolved as well, as the file detail will no longer exist

> Maybe these issues are for another task. And maybe it's just me who's confused.
> Also yes, there's no need to reload resources on resource creation since the resource list loads them automatically on mount. I suspect this is a remnant of the tree component use.

MichalMed (approved):
I'm not really sure if it works as expected or not. 
1) While creating resources, I'm able to create "resource" or "document". What exactly is resource other than document or file in TermIt?
2) if I create Document and attach a file to it, I cannot see it in resources list, however I see the older files. I expect the visible documents are due to deprecation and newly the access to the files will be edited via https://kbss.felk.cvut.cz/redmine/issues/1379. 
3) I agree with Martin that it after deleting file, user shall be redirected to the document, not the list of resources.

psiotwo:
> 1. While creating resources, I'm able to create "resource" or "document". What exactly is resource other than document or file in TermIt?

resource = "something that can be tagged with TermIt concepts" (e.g. a document content of which is not in TermIt) - resources were meant to be annotated by TermIt concepts so that they can be searched for (e.g. in LKOD, but we might plan this functionality directly in TermIt as well)

> 2. if I create Document and attach a file to it, I cannot see it in resources list, however I see the older files. I expect the visible documents are due to deprecation and newly the access to the files will be edited via https://kbss.felk.cvut.cz/redmine/issues/1379.

"I cannot see it in resources list," - You mean you cannot see the file but you can see the document, right? 

Also, https://kbss.felk.cvut.cz/redmine/issues/1379 does not seem to deal with editing - but in the resources menu you should be able to see just documents and going to a document detail, you will see its filed and will be able to manage them there.

> 3. I agree with Martin that it after deleting file, user shall be redirected to the document, not the list of resources.

File deletion will only be possible from document detail (not from file detail, because it won't exist)

PR #77, opened 2021-01-15
[R1444] review feedback

psiotwo:


MichalMed (approved):
It looks OK, however it seems that the label of "Komentář" was changed to "Doplňující poznámka". AFAIK we did not discuss that, but it's OK.

PR #78, opened 2021-01-18
[R1378] document attach/detach from vocabulary

psiotwo:
Done
- attaching/detaching an existing document from the vocabulary screen

TODO
- creating a new document at the vocabulary screen
- attaching/detaching an existing vocabulary to a document
- tests


MichalMed (changes requested):
What happens to the document if I deatach it? I tried on the "very new test vocab", which was created as document vocabulary, then I have deatached the document and tried to attach it again, but I have not found it in the list of existing documents. I would also expect a possibility to create a new document in the vocabulary, that does not contain any. I have created a new document test document 1378 and attached it via select existing document from . It worked, but I have got a red error (cannot access server). I still do not know, what happened to the original document, that was created along with the vocabulary after deataching.

ledsoft:
I'm not a fan of adding yet another action button to the vocabulary detail header. What about having the Document tab there at all times and if there is no document, have an info message with a button triggering the document attachment?

ledsoft:
Also, how are attachable documents resolved? I see 58 documents in TermIt, but only 18 of them are available for attaching.

psiotwo:
> Also, how are attachable documents resolved? I see 58 documents in TermIt, but only 18 of them are available for attaching.

You can see only those that are not attached to any vocabulary yet.

ledsoft:
Ok, testing functionality:
I selected a document (the selection dialog is weird IMHO, at least the cursor should be a pointer to indicate that the rows are actionable) but nothing happened (document tab did not appear, just the dialog was closed). So I refreshed the page, still no document tab. But when I tried opening the document selection dialog, the document I selected previously is no longer there.
Repository shows that the document has been attached, so I guess this might be a caching issue?

psiotwo:
- Added the option to create a document in vocabulary detail.
- Moved the Attach document/Detach document actions to the document tab.
- Changed cursor in document list to pointer.

MichalMed (changes requested):
Moving buttons to Document tab is great. I can also see detached documents among all documents and I'm able to attach it again, or to another vocabulary.
BUT: if I detach document and see the documents detail, there is still Document vocabulary filled. Now I'm not sure, if this is bug or feature. Anyway, it only happens for the document that is created automatically along with the vocabulary. If I delete this one and create a new one attached to the same vocabulary, then detach it, this does not happen.

psiotwo:
> Moving buttons to Document tab is great. I can also see detached documents among all documents and I'm able to attach it again, or to another vocabulary.
> BUT: if I detach document and see the documents detail, there is still Document vocabulary filled. Now I'm not sure, if this is bug or feature. Anyway, it only happens for the document that is created automatically along with the vocabulary. If I delete this one and create a new one attached to the same vocabulary, then detach it, this does not happen.

Fixed. Also changed the 'Remove File' icon color to Red

MichalMed (approved):
lgtm

PR #79, opened 2021-01-19
redesign validation results visualization

psiotwo:


MichalMed (approved):
it looks much better than before.

PR #80, opened 2021-01-21
1397 definition block in term detail

ledsoft:
https://kbss.felk.cvut.cz/redmine/issues/1397


psiotwo:
Having the document inside the definition block is quite confusing to me
(the definition block as a whole is confusing to me both in edit and view mode as well, but it has already been discussed)

ledsoft:
This is according to task requirements.

psiotwo:
> This is according to task requirements.

OK, the former confusion comes from the "Dokument" label, which comparing to "Zdroj" does not make much sense to me. Also, even if it stays like that the button "Zobrazit zdroj" is confusing as well, because the field is now called "Dokument" ...

blcham:
Visualization is for me confusing as well. I guess it requires rather an indentation of fields Text, Zdroj, Dokument then separators.

"Dokument" label is for me confusing as well. I would expect to see something like "Location in document" ...

Moreover, it might be confusing that the user does not have a chance to find out what is "Zdroj". He will remember that he can find out more info in edit mode, so he switches to edit mode, but "Zdroj" attribute is not there. I suggest showing some info on hover over the text Zdroj at least. An of course same for other attributes.

ledsoft:
Ok, what would be your suggestions for the "Dokument" vs "Zdroj" labels? "Zdroj" corresponds to dc:source.

W.r.t. no "Zdroj" in edit mode, that was an incorrect label, I fixed it now.

MichalMed (approved):
I like it. Question is how to call the link to definition in document. Originally it was Source of definition. Maybe we could call it "Definition in document" and replace "Go to source" by "Go to definition in document"

ledsoft:
I would prefer something shorter, even if just for the sake of the layout.

blcham:
So how about "Location in document" as I proposed earlier ...

Anyway, for me, there is only one "Zdroj" (source of the definition).  If you had this discussion, or I interpret it incorrectly, just ignore the rest of the text. To me "external link" or "internal link" are only attributes of  "Zdroj". So I would visualize it as one form field "Zdroj" with two "separate answers" (with proper visualization, e.g. next to each other or one above the other).

Structurally, i mean:
-- term
---- Definice
      ------ Zdroj
           -------- Internal link
           -------- External link

ledsoft:
Great, "Location in document" is much shorter than "Definition in document" :D

ledsoft:
Otherwise I partially agree than we could visualize it as "source" and it would be either clickable internal link, leading to the annotator or external link/not clickable in case no document in TermIt is available for it.
The only downside to this is that now you can clearly tell the file in which the definition is before clicking on it. If we visualize it as the dc:source value, which would somehow lead into the annotator, the user looses this information. If you get what I mean.

blcham:
> Great, "Location in document" is much shorter than "Definition in document" :D

I knew you would like it! It is 2 chars less :):):)

blcham:
> Otherwise I partially agree than we could visualize it as "source" and it would be either clickable internal link, leading to the annotator or external link/not clickable in case no document in TermIt is available for it.
> The only downside to this is that now you can clearly tell the file in which the definition is before clicking on it. If we visualize it as the dc:source value, which would somehow lead into the annotator, the user looses this information. If you get what I mean.

1) I vaguely recall Michal mentioning that he needs both internal and external links at the same time.
2) I don't understand why we have to remove the text "Soubor pro slovnik MPP..." ?

Could it look like this ?:
Selection_349

blcham:
OK, Pozice is maybe stupid word, but you know what i mean ... :)

psiotwo:
IMO
1. Any of "Go to definition in document", "See in the document", or even "Go there" is a better replacement for "Go to source".
2. Another option is to remove "Document" field name, then the internal link would appear as another row of a source. I am against "Pozice v" - the less text will be there the better seems to me.

I am OK with either of these options.

ledsoft:
I like the second option. Will implement it and we'll see.

blcham:
> 2. Another option is to remove "Document" field name, then the internal link would appear as another row of a source.

Thats what i meant by picture :)

ledsoft:
Ok, I updated the UI based on @blcham / @psiotwo  proposals.

psiotwo:
> Ok, I updated the UI based on @blcham / @psiotwo proposals.

I am OK with this

psiotwo:
> > Ok, I updated the UI based on @blcham / @psiotwo proposals.
> 
> I am OK with this

maybe not part of this ticket, but I would also remove the word "Text"

ledsoft:
> maybe not part of this ticket, but I would also remove the word "Text"

Any particular replacement? Or just remove it completely.

blcham:
Looks good to me. 

Although the minor issue is this visualization of the section as I mentioned. Now we have 2 attributes only and @psiotwo does not like the word "Text". So how about removing section definition and having only two flat attributes:
Definice:
Zdroj definice:

ledsoft:
Well, then we're back where TermIt started before this task, only slightly reordered attributes.

blcham:
The small change at the end, but it seems to me much more clear with merged attributes together.

Sorry for iterating over source code changes ... I prefer this flat way, but as I said I am ok with both.

psiotwo:
> > maybe not part of this ticket, but I would also remove the word "Text"
> 
> Any particular replacement? Or just remove it completely.

No replacement. IMO, the definition is expected to be text by users and we do not need to "surprise" him/her that it is indeed the case ...

ledsoft:
Ok, to prevent pushing unnecessary commits. Here is the current version:
definition-with-labels

And here is a version without labels:
defintion-no-labels

For me, the version without labels is confusing, there is no telling which part of the definition is the text itself and which is source identification.

psiotwo:
> Ok, to prevent pushing unnecessary commits. Here is the current version:
> definition-with-labels
> 
> And here is a version without labels:
> defintion-no-labels
> 
> For me, the version without labels is confusing, there is no telling which part of the definition is the text itself and which is source identification.

For me it is confusing as well. As I wrote, I would prefer to keep "source" and remove "text"

ledsoft:
Like this?
definition-no-text-label

ledsoft:
Ok, I've modified the definition block to the last featured layout. I'll still wait for @MichalMed to check it out as well before merging. Thanks for the input.

ledsoft:
Another version based on today's discussion:
definition-block

MichalMed:
It's a little gritty, but so far the best solution.

ledsoft:
Ok, I'll update the styling of resource and vocabulary detail based on the changes on term detail and merge this PR afterwards. Thanks

PR #83, opened 2021-01-26
[R1378] termassignmentsinfo for document

psiotwo:
https://kbss.felk.cvut.cz/redmine/issues/1378:
- assigned terms field for document now shows assigned terms for the document itself plus assigned terms for all its files

UI change proposals:
- changed color of "Detach document" button to 'muted'
- simplified file list table (removed header, renamed header and the add button)


MichalMed (changes requested):
From the https://kbss.felk.cvut.cz/redmine/issues/1380:
* In the Document detail shall also be blocked “Delete” button, which gets unblocked only if there are no Files attached to the Document. This shall also be written somewhere next to the blocked button.
* If the Document is not attached to any Vocabulary, there shall be a button allowing the user to attach it to the Vocabulary.

Those requirements are not fulfilled. 
To the first bullet - I imagined the delete button will be present all the time, but not clickable, or after click, there appear a popup with "The document cannot be deleted, because it contains at least a file."
Second bullet is not fulfilled at all, but it was not part of this PR.

psiotwo:
Fixed the delete button. The second bullet won't be implemented now.

MichalMed (approved):
lgbt

PR #87, opened 2021-01-30
[R1476] removed IRI editing when creating new file

psiotwo:
hotfix

### DIFF
--- a/src/component/resource/ResourceMetadata.tsx
+++ b/src/component/resource/ResourceMetadata.tsx
@@ -2,7 +2,7 @@ import * as React from "react";
 import {injectIntl} from "react-intl";
 import classNames from "classnames";
 import withI18n, {HasI18n} from "../hoc/withI18n";
-import {CardBody, Col, Label, Row, Table} from "reactstrap";
+import {CardBody, Col, Label, Row} from "reactstrap";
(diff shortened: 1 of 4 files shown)


src/component/resource/ResourceMetadata.tsx
```diff
@@ -16,23 +16,24 @@ interface ResourceMetadataProps extends HasI18n {
...
-            <Row>
-                <Col xl={2} md={4}>
-                    <Label className="attribute-label mb-3">{props.i18n("resource.metadata.description")}</Label>
-                </Col>
-                <Col xl={10} md={8}>
-                    <p id="resource-metadata-description">{(props.resource || {}).description}</p>
-                </Col>
-            </Row>
-            {props.additionalColumns}
+            <Table>
```
ledsoft: Why is there a table here? Especially given no rows/cells are used in it.

ledsoft (commented):
See the comment. In addition:
- The new file dialog does not allow me submit it without attaching a physical file, even though I have filled a label and the submit button is enabled. Either the submit button should be disabled until a file is attached, or it should be possible to submit the dialog without attaching a physical file.

ledsoft (approved):
LGTM

PR #90, opened 2021-02-08
1480 public view on unauthorized tests

ledsoft:
https://kbss.felk.cvut.cz/redmine/issues/1480


psiotwo (commented):
I can will take closer look once at comp, but why are there 50 changed files?

ledsoft:
Synchronization of branches.

psiotwo (approved):
lgtm

PR #92, opened 2021-02-08
[Fix] Ensure registration link is not shown in the public view when o…

ledsoft:
…nly admin-based user registration is enabled.


psiotwo (approved):
lgtm

PR #94, opened 2021-02-11
[Bug #1492] Show tree select value as internal link where relevant.

ledsoft:
https://kbss.felk.cvut.cz/redmine/issues/1492

Tree select value is shown as internal link for:
- Vocabulary imports edit
- Term parent selector
- Term selector in annotator

There are two more occurrences of the tree select: 
- term type selector - has been left as is, so the link goes to the type IRI
- term selector in term detail - displays value as simple (non-clickable) label, since the user is already viewing the detail of the selected term.

### DIFF
--- a/config/server.json
+++ b/config/server.json
@@ -1,3 +1,3 @@
 {
-  "url": "https://kbss.felk.cvut.cz/termit-server-dev"
+  "url": "TO-BE-PROVIDED"
 }
--- a/config/server.json
+++ b/config/server.json
@@ -1,3 +1,3 @@
 {
-  "url": "TO-BE-PROVIDED"
+  "url": "https://kbss.felk.cvut.cz/termit-server-dev"
 }
--- a/package-lock.json
+++ b/package-lock.json
@@ -1694,9 +1694,9 @@
       }
     },
     "@babel/runtime-corejs2": {
-      "version": "7.12.5",
-      "resolved": "https://registry.npmjs.org/@babel/runtime-corejs2/-/runtime-corejs2-7.12.5.tgz",
-      "integrity": "sha512-kt5YpZ7F5A05LOgQuaMXXmcxakK/qttf5C/E1BJPA3Kf5PanbjPzDoXN+PIslUnjUxpuKblCsXyP0QfMiqyKqA==",
+      "version": "7.12.13",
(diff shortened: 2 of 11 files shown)


config/server.json
```diff
@@ -1,3 +1,3 @@
 {
-  "url": "TO-BE-PROVIDED"
+  "url": "https://kbss.felk.cvut.cz/termit-server-dev"
```
psiotwo: why this value goes to git?

psiotwo (changes requested):
> https://kbss.felk.cvut.cz/redmine/issues/1492
> 
> Tree select value is shown as internal link for:
> 
> * Vocabulary imports edit
> * Term parent selector
> * Term selector in annotator

- term parent selector - works ok
- vocabulary imports edit - for 'ML Test' I can see two imported vocabularies, but the edit screen does not show them selected and the tree select is empty
- term selector in annotator - works ok

> There are two more occurrences of the tree select:
> 
> * term type selector - has been left as is, so the link goes to the type IRI
- this should probably be replaced by the copy-link functionality to keep the UI consistent. At present, I propose to remove the link altogether.

config/server.json
ledsoft: This is of course an accidental change. Will fix it in the morning.

ledsoft:
@psiotwo 
- ML test vocabulary imports is a data issue. Someone deleted vocabulary identified by `http://onto.fel.cvut.cz/ontologies/slovnik/slovnika` and didn't do a proper cleanup, so ML test ended up pointing to a nonexistent vocabulary
- Regarding type selector. Copy link functionality would keep the UI consistent in what sense? Consistent with what?

psiotwo:
With the other external links (terms/vocs/resources)

ledsoft:
Ok, but that is not copy functionality, that is external link. Will make corresponding changes.

psiotwo:
> Ok, but that is not copy functionality, that is external link. Will make corresponding changes.

Ah true, in lists we have the external link, I mismatched it with the detail screens which have the copy-link.

psiotwo (approved):
lgtm

MichalMed:
I'm just not sure what do you mean by term selector in term detail. Otherwise it works as expected.

ledsoft:
The tree select on the right.

PR #98, opened 2021-02-25
1488 authorization

ledsoft:
https://kbss.felk.cvut.cz/redmine/issues/1488


src/component/hoc/withLoading.tsx
```diff
@@ -4,7 +4,7 @@ import getDisplayName from "../../util/getDisplayName";
 import classNames from "classnames";
 
 export interface WithLoadingProps {
-    loading: boolean,
+    loading?: boolean,
```
psiotwo, line 7: why loading is optional?

src/i18n/en.ts
```diff
@@ -499,6 +500,7 @@ const en = {
         "annotator.legend.definition.tooltip": "Definition of a known term created by a user.",
         'annotator.legend.toggle.show': 'Show legend',
```
psiotwo: You might want to replace single quotes here as well

psiotwo (approved):
lgtm, minor comment

src/component/hoc/withLoading.tsx
ledsoft: The HOC API was causing typing issues. The effect is the same so it was simpler than trying to figure out how the API should look like...

src/i18n/en.ts
ledsoft: Good point. Should simplify merging to KODI.

PR #102, opened 2021-03-03
1510 public view definition not shown

ledsoft:
https://kbss.felk.cvut.cz/redmine/issues/1510

Note that this PR requires https://github.com/kbss-cvut/termit/pull/46 to be merged and deployed as well, because it loads configuration in public view, which is not supported by the currently deployed version.


psiotwo (approved):
lgtm

PR #106, opened 2021-03-10
Refactoring/hooks

ledsoft:
The PR is massive, but the changes are quite simple. I created a custom hook for i18n and replaced the use of the
`injectIntl(withI18n(...))` HOC pattern with this new hook. This greatly simplifies the component structure in the page.


psiotwo (approved):
lgtm, will simplify code

PR #127, opened 2021-03-30
[Feature #1544] Show vocabulary which was used to annotate file

### LINKED ISSUE #123: Anotátor by měl zobrazovat slovník, který používá
# Popis
Jako **editor slovníku**
potřebuji **vidět, který slovník používá anotátor pro textovou analýzu**
abych **mohl dokončit editaci pojmů původně vytvořených v anotátoru v detailu slovníku. **

Odkaz na slovník použitý anotátorem bude klikatelný a bude vést na detail slovníku v TermIt.

ledsoft:
https://kbss.felk.cvut.cz/redmine/issues/1544

### DIFF
--- a/src/component/annotator/Annotator.tsx
+++ b/src/component/annotator/Annotator.tsx
@@ -6,7 +6,7 @@ import Term from "../../model/Term";
 import HtmlDomUtils from "./HtmlDomUtils";
 import LegendToggle from "./LegendToggle";
 import {DomUtils} from "htmlparser2";
-import VocabularyUtils, {IRI} from "../../util/VocabularyUtils";
+import VocabularyUtils, {IRI, IRIImpl} from "../../util/VocabularyUtils";
 import CreateTermFromAnnotation, {CreateTermFromAnnotation as CT} from "./CreateTermFromAnnotation";
 import SelectionPurposeDialog from "./SelectionPurposeDialog";
 import {connect} from "react-redux";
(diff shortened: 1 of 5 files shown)


src/component/annotator/Annotator.tsx
```diff
@@ -387,35 +392,56 @@ export class Annotator extends React.Component<AnnotatorProps, AnnotatorState> {
...
-            <div id="annotator"
-                 ref={this.containerElement}
-                 onMouseUp={this.handleMouseUp}>
-                <AnnotatorContent content={this.state.internalHtml} prefixMap={this.state.prefixMap}
-                                  stickyAnnotationId={this.state.stickyAnnotationId}
-                                  onCreateTerm={this.onCreateTerm} onUpdate={this.onAnnotationTermSelected}
-                                  onRemove={this.onRemove} onResetSticky={this.resetStickyAnnotationId}/>
-            </div>
-        </div>
+            <HeaderWithActions title={this.props.i18n("annotator") + " - " + this.props.file.getLabel()}/>
```
psiotwo: Why to prefix the title with "annotator - "?

psiotwo (approved):
Technical-wise LGTM.

As I user I don't like this solution much - in addition to the inline comment, I would prefer having the vocabulary info not distracting me when scanning the text, e.g. aligned to the right.

ledsoft:
If the vocabulary link is moved to the right, should there also be the attribute label ("Používá pojmy ze slovníku"). If not, would it not be confusing as to what does the vocabulary mean? If yes, how to align them?

psiotwo:
> If the vocabulary link is moved to the right, should there also be the attribute label ("Používá pojmy ze slovníku"). If not, would it not be confusing as to what does the vocabulary mean? If yes, how to align them?

Actually, the label is confusing - the file does not necessarily use any terms from the vocabulary. The information is useful mainly for the Analysis action itself. User can just assume that the current annotations in the document were suggested using the vocabulary - they might have been manually crafted, for example. 

Thus, the vocabulary might be closer to the Analysis button and  it might be changeable (but it is another ticket)

ledsoft:
Ok, removed the label and moved the link to the right.

PR #129, opened 2021-03-30
1545 analyze select vocabulary

ledsoft:
https://kbss.felk.cvut.cz/redmine/issues/1545


psiotwo (commented):
It is a bit annoying and confusing that for a file not attached to a vocabulary I am asked twice for the vocabulary that should be used for text analysis. Otherwise LGTM

ledsoft:
As I said on our call. That would require significant changes in the annotator. Moreover, that will happen only before running text analysis for the first time.

psiotwo (approved):
lgtm,
if possible, please adjust texts of the two dialogs not to confuse users:
- the first one is for manual annotation
- the second one is for automatic text analysis

PR #139, opened 2021-04-15
1288 form layout

ledsoft:
https://kbss.felk.cvut.cz/redmine/issues/1288


psiotwo:
deploy preview does not seem to work - maybe wrong server URL?

ledsoft:
This is apparently caused by the switch to CRA, which means different environmental variables are now used. I updated all the available configs (including Docker) but apparently the Netlify build has to be configured separately. I don't have sufficient rights to do that. @psiotwo

psiotwo:
> This is apparently caused by the switch to CRA, which means different environmental variables are now used. I updated all the available configs (including Docker) but apparently the Netlify build has to be configured separately. I don't have sufficient rights to do that. @psiotwo

fixed

psiotwo:
Comments:
- I never liked the horizontal bars around definition, but now I don't like them even more as they distract me without any apparent value. Would propose to rename Definice -> "Text" to "Text definice" or "Definice - Text" and Definice -> "Zdroj" to "Zdroj definice" or "Definice - Zdroj"
- The icon for symbolizing multi-linguality might be confusing for our (Japanese?) friends :-). Would choose something language-independent.
- Text "povinné" is too far from the corresponding input. Either move it closer, or I would propose to even move it to the label (e.g. "Název (povinný)" or "Název *"
image

ledsoft:
1. That would be up to discussion with @MichalMed , who proposed the definition attributes to be in a block binding them together (and I agree with him). We can discuss how else to visualize this block.
2. It is a commonly used translation icon (Chrome uses it as well). What language-independent icon would you propose?
3. What browser, which page, what screen resolution did you have when taking the screenshot? For me, in Chrome as well as Firefox, it shows directly under the input.

image

psiotwo:
> 2. It is a commonly used translation icon (Chrome uses it as well). What language-independent icon would you propose?

Maybe some globe icon you proposed earlier seems a good fit here to me, e.g. https://www.iconfinder.com/icons/3554052/earth_globe_internet_language_language_switch_planet_translate_icon

But I don't care much about this, it was more teasing from my side than a real issue.

> 3. What browser, which page, what screen resolution did you have when taking the screenshot? For me, in Chrome as well as Firefox, it shows directly under the input.

Page: https://deploy-preview-139--termit-dev.netlify.app/#/vocabularies/ml-test/terms/dokumentace?namespace=http://onto.fel.cvut.cz/ontologies/slovnik/
Resolution: any resolution I tried
Browser: Chrome 89.0.4389.128 (Official Build) snap (64-bit)

src/component/misc/HelpIcon.tsx
```diff
@@ -0,0 +1,55 @@
...
+  const [pinned, setPinned] = React.useState(false);
+  const show = () => setOpen(true);
+  const mouseOut = () => {
+    if (!pinned) {
+      setOpen(false);
+    }
+  };
+  const onClick = () => {
+    if (!pinned) {
+      setPinned(true);
```
psiotwo: setPinned(!pinned)
setOpen(!pinned)

psiotwo (approved):
lgtm, see comment

psiotwo (approved):
lgtm

MichalMed (approved):
to the conversation you had previously

1. definition must remain as a block, that is important.
2. I'm disgusted by the new icon with globe, the old one was much better. Actually the first thing I've thought of when I saw the whole page for the first time was: "what is the globe for?!?" Is it neseccary at all? The procesds of term ceratin shall be fill in everything in one language, then in second language. Elements not supported will remain the same, so why force it?
3. i see it close to the label, too (both in firefox and chromium)

Otherwise LGTM

psiotwo:
> to the conversation you had previously
> 
> 1. definition must remain as a block, that is important.

Let's keep it. 

> 2. I'm disgusted by the new icon with globe, the old one was much better. Actually the first thing I've thought of when I saw the whole page for the first time was: "what is the globe for?!?" Is it neseccary at all? The procesds of term ceratin shall be fill in everything in one language, then in second language. Elements not supported will remain the same, so why force it?

Sure, as I wrote, I am OK with any icon. If the user does not know that an attribute is not multilingual, he will be confused, because two same-looking text fields will behave differently after switching to another language. Would look like an error of the app.

PR #140, opened 2021-04-15
[Support #1563] Minor annotator UI fixes based on user feedback.

ledsoft:
https://kbss.felk.cvut.cz/redmine/issues/1563


psiotwo:
Minor suggestion:
- "Pojem" appears there twice
image

ledsoft:
So I guess we're not using the same button label as in vocabulary as we discussed previously.

psiotwo:
> So I guess we're not using the same button label as in vocabulary as we discussed previously.

Yes, I am aware of this mismatch. Another option would be to remove "Pojem:"  label (as the header might already be intuitive enough - "Výskyt pojmu"). I don't care much which option to pick (or keeping the original version). 

@MichalMed might have a stronger opinion here.

MichalMed:
Hi,
First, I'm not able to get to netlify deployment. After few seconds I'm getting blank screen.
Second, I'd rather keep label and shorten the label to the right of "+" to simple "+ Nový". It would look like this:

Pojem: [ + Nový]

which is IMHO ok.

ledsoft:
KBSS is apparently inaccessible, will ping you once it is back.

And yes, I went with the solution you mention.

MichalMed (changes requested):
- I like the + New button,
- color of (x) is fixed,
- text what is the purpose.. was replaced, nice. Just a comment, the new text is longer and it transforms the shape of the whole popup (see the picture),
- autofocus does not work (both firefox and chromium).
Screenshot from 2021-04-19 18-52-41

ledsoft:
@MichalMed Thanks for the feedback. Will modify the popover buttons sizing.
As for the focus issue - I disabled it because it caused major issues with scrolling. Basically, the screen would jump up to the top of the document any time the input were supposed to autofocus. I suspect the issue is in the interplay of the popover, its absolute position and the focus event. I wasn't able to get around it...

MichalMed:
In that case, merge it as it is

ledsoft:
@MichalMed I managed to get the auto focus working. Please, check it out again and approve (if it's ok).

MichalMed (approved):

PR #147, opened 2021-04-26
1562 show source in file

ledsoft:
https://kbss.felk.cvut.cz/redmine/issues/1562


MichalMed (approved):
I'm not sure what it shall do. Basically, I can see the button now even if there is no source, which is probably purpose of this ticket. In that case, it works and I approve it.
However, if I click the button, I do not get anywhere because I get the error described in redmine issue (1562). It is still quite difficult to even find a piece of text in the annotator that can be marked as definition :rofl:

ledsoft:
Yeah, I replied to your description in the ticket - it is a separate issue for which I created a new ticket: https://kbss.felk.cvut.cz/redmine/issues/1585

PR #149, opened 2021-04-29
1585 term definition source link bug

ledsoft:
https://kbss.felk.cvut.cz/redmine/issues/1585


MichalMed (commented):
I tried the feature on term "stavební pozemek" from Building Act and I've got this screen:
Screenshot from 2021-04-30 12-49-40
I suppose it is again another issue, but it practically does not allow users to make it work.

Then I've tried that for term "Targeted character of locality" from Metropolitan plan. I was redirected to the text of a file with message "I'm not able to highlight the definition"

ledsoft:
This is a data issue. There is no vocabulary (or text analysis record) associated with the file being opened.

I suspect this was caused by marking the definition of the term in the file and then disconnecting the document from the corresponding vocabulary (as evidenced by the fact that the vocabulary in question has no assigned document). This causes the annotator not to be able to determine which vocabulary to use for annotation.

MichalMed:
I understand that, however, it has no influence on showing the definition in a file. This message shall be shown at the moment when I try to annotate content

ledsoft:
The current annotator requires the vocabulary to be known before opening the file content. Changing this behavior would require SIGNIFICANT changes to its implementation.

MichalMed (approved):

PR #68, opened 2020-12-22
[R1377] implementation

psiotwo:
- https://kbss.felk.cvut.cz/redmine/issues/1377
- refactoring of Advanced sections

PR #150, opened 2021-04-30
[R-#1580] exactMatches

psiotwo:
https://kbss.felk.cvut.cz/redmine/issues/1580

A corresponding server adjustments are needed - https://github.com/kbss-cvut/termit/pull/68


src/action/AsyncActions.ts
```diff
@@ -836,17 +836,51 @@ export function searchTerms(searchString: string) {
     };
 }
 
+export function loadAllTerms(
```
ledsoft, line 798: Why do you declare a new action (and the corresponding REST endpoint and backend stack) and not use the term fetching logic used for resource term assignment? It does pretty much the same thing, does it not?

psiotwo: I decided to keep both endpoints for now, as their semantics is different:
1) /search/fts for the fulltext search (both terms and vocabularies)
2) /terms/roots for paginated requests in autocomplete inputs during term editing.

Nevertheless, I have harmonized their usage - ResourceTermAssignmentsEdit now uses the /terms/roots endpoint, leaving /search/fts to the top bar full-text search only.

ledsoft: I didn't imply the endpoints should be removed, I was merely questioning the need for two APIs for the same feature. Clearly we don't want to remove FTS.

src/model/Term.ts
```diff
@@ -180,12 +172,14 @@ export default class Term extends Asset implements TermData {
...
-                pt.toTermData(true)
-            );
+        if (result.exactMatchTerms) {
+            result.exactMatchTerms = result.exactMatchTerms.map( (em : Term) =>
+            { return {
+                types: em.types,
+                iri : em.iri,
+                label: em.label,
+                vocabulary: em.vocabulary
+            }}) ;
```
ledsoft: `Term.exactMatchTerms` are (or at least should be based on the class declaration) instances of `TermInfo`, not `Term`, so I don't see any reason for this explicit processing.

src/component/term/TermMetadataEdit.tsx
```diff
@@ -150,6 +151,12 @@ export class TermMetadataEdit extends React.Component<
     this.setState({ parentTerms });
   };
 
+  public onExactMatchesChange = (exactMatchTerms?: Term[]) => {
+    this.setState({
+      exactMatchTerms: exactMatchTerms?.map((e) => e as TermInfo),
```
ledsoft, line 156: Ok, now I probably see why there was the transformation routine for exactMatch attribute in `Term.toJsonLd`. In that case, I think the selected terms should be mapped to `TermInfo` here - i.e., create new objects with just the subset of attributes corresponding to `TermInfo`.

ledsoft (approved):
LGTM, see the comment

PR #152, opened 2021-05-05
[#109] fixed badge

psiotwo:
https://github.com/kbss-cvut/termit-ui/issues/109
(redmine has old data at present, so bypassing)


MichalMed (approved):
functional, BUT
I think that blank validation tab is confusing anyway, may be there shall be something like "no valůidation errors" or something

psiotwo:
> functional, BUT
> I think that blank validation tab is confusing anyway, may be there shall be something like "no valůidation errors" or something

Good point, fixed.

PR #155, opened 2021-05-08
1549 skos related

ledsoft:
https://kbss.felk.cvut.cz/redmine/issues/1549

Note that this PR requires backend version from https://github.com/kbss-cvut/termit/pull/72 to be running on the server.


psiotwo (approved):
lgtm

PR #158, opened 2021-05-12
[R-#1555] minor label changes in file creation

psiotwo:
https://kbss.felk.cvut.cz/redmine/issues/1555


src/i18n/cs.ts
```diff
@@ -511,6 +511,7 @@ const cs = {
             'Soubor "{fileName}" úspěšně nahrán na server.',
         "file.annotate.selectVocabulary":
             "Nelze určit slovník pro anotování tohoto souboru. Vyberte ho, prosím...",
+        "file.upload": "Nahraj",
```
MichalMed: i suggest to replace with infinitive "Nahrát"

MichalMed (changes requested):
one comment

PR #160, opened 2021-05-13
1571 annotator layout

ledsoft:
https://kbss.felk.cvut.cz/redmine/issues/1571

Based on feedback

### DIFF
--- a/src/i18n/en.ts
+++ b/src/i18n/en.ts
@@ -369,7 +369,8 @@ const en = {
         "term.metadata.altLabels.placeholder":
             "Enter a new synonym and press 'Add'",
         "term.metadata.altLabels.addButton.text": "Add",
-        "term.metadata.altLabels.addButton.title": "Click to add the alternative label value",
+        "term.metadata.altLabels.addButton.title":
+            "Click to add the alternative label value",
         "term.metadata.altLabels.remove": "Remove synonym",
         "term.metadata.altLabels.remove.text": "Remove",
         "term.metadata.altLabels.help":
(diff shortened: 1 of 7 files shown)


src/i18n/en.ts
```diff
@@ -378,7 +380,9 @@ const en = {
         "term.metadata.hiddenLabels.label": "Search strings",
         "term.metadata.hiddenLabels.placeholder":
             "Enter a new search string and press 'Add'",
-        "term.metadata.hiddenLabels.placeholder.text": "Add",
+        "term.metadata.hiddenLabels.addButton.text": "Add",
+        "term.metadata.hiddenLabels.addButton.title":
+            "Click to add the hidden label value",
```
psiotwo: Why not simply "Click to add the search string" (also, the UI speaks about search strings, so it should be consistent)

src/i18n/en.ts
```diff
@@ -368,7 +368,9 @@ const en = {
         "term.metadata.altLabels.label": "Synonyms",
         "term.metadata.altLabels.placeholder":
             "Enter a new synonym and press 'Add'",
-        "term.metadata.altLabels.placeholder.text": "Add",
+        "term.metadata.altLabels.addButton.text": "Add",
+        "term.metadata.altLabels.addButton.title":
+            "Click to add the alternative label value",
```
psiotwo: Why not simply "Click to add the synonym" (also, the UI speaks about synonyms, so it should be consistent)

psiotwo (commented):
see comments

psiotwo (approved):
lgtm

PR #169, opened 2021-05-27
1591 navigation between term with same label

ledsoft:
https://kbss.felk.cvut.cz/redmine/issues/1591

Can be tested on term "Lokalita" in vocabulary "ML test" on dev.


psiotwo (approved):
lgtm

PR #171, opened 2021-06-03
[Bug #1587] Ensure subsequent navigation to the same file works correctly

ledsoft:
https://kbss.felk.cvut.cz/redmine/issues/1587

This fixes the issue where repeated navigation to the same file (e.g. via the term definition source link) was ending with a  blank annotator screen. It was caused by the browser (Chrome) automatically encoding location.href, which is used by the `ResourceFileDetail` component to determine whether to reloaded the content or not.


MichalMed (approved):
works

psiotwo (approved):
lgtm

PR #172, opened 2021-06-07
1564 annotator selection extension

ledsoft:
https://kbss.felk.cvut.cz/redmine/issues/1564
https://kbss.felk.cvut.cz/redmine/issues/1293

Also #121


MichalMed (approved):
temporary solution, but works well

PR #173, opened 2021-06-08
1419 skos import

psiotwo:
- [R-#1419](https://kbss.felk.cvut.cz/redmine/issues/1419)


MichalMed:
after importing skos, i get unspecified red error in upper right corner and get redirected to vocabularies list

psiotwo:
@MichalMed backend deployed to dev.

MichalMed:
na devu nic není

MichalMed (changes requested):
Vocabulary may be imported. After uploading the file and pressing import, I get the message that everything went well, but I:
1) do not know the name of the vocabulary,
2) can not find it in the vocabularies list,
3) can not find terms from the "imported" vocabulary in TermIt.

Basically I suspect the system that it only tells me it is OK, but actually nothing happened.

MichalMed (approved):
lgtm

PR #175, opened 2021-06-18
1558 term validation message ux

ledsoft:
https://kbss.felk.cvut.cz/redmine/issues/1558

@psiotwo I introduced a new property to inputs - `validationMessage` which represents message from the quality validation. However, I don't like to similarity with the `invalidMessage` property, which represents the message to show when the input value is really invalid and the form cannot be submitted. Do you have any better ideas for naming the property?


psiotwo:
> https://kbss.felk.cvut.cz/redmine/issues/1558
> 
> @psiotwo I introduced a new property to inputs - `validationMessage` which represents message from the quality validation. However, I don't like to similarity with the `invalidMessage` property, which represents the message to show when the input value is really invalid and the form cannot be submitted. Do you have any better ideas for naming the property?

There is already "invalid?" flag. So why not to have just a single message field?
I think we should merge all validation functionalities into a single one in the future, so I would think we should visualize it in a uniform way in the UI.

ledsoft:
Well, according to https://kbss.felk.cvut.cz/redmine/issues/1558 we probably should not. One thing is the validation of the form which indicates whether it can be submitted at all (that's the invalid flag) and another is the term quality validation which does not prevent the user from saving the data.

psiotwo:
> Well, according to https://kbss.felk.cvut.cz/redmine/issues/1558 we probably should not. One thing is the validation of the form which indicates whether it can be submitted at all (that's the invalid flag) and another is the term quality validation which does not prevent the user from saving the data.

I expected there would be one or more validation messages with severities (no matter which source they come from). If a single message has the severity of an error, it is blocking the submission and the 'invalid' flag must be true. Based on it, the UI marks the widget as submittable or not and renders the messages with their severities (probably color-based).

ledsoft:
@psiotwo reimplemented

src/component/misc/validation/InputValidationMessage.scss
```diff
@@ -0,0 +1,10 @@
+$qualityAffectingRuleViolationColor: #8965e0;
```
psiotwo, line 1: Why not yellow variants are used here?

src/component/misc/validation/InputValidationMessage.tsx
```diff
@@ -0,0 +1,30 @@
...
+interface InputValidationMessageProps {
+  message: ValidationResult;
+}
+
+const InputValidationMessage: React.FC<InputValidationMessageProps> = (
+  props
+) => {
+  const { message } = props;
+  const { i18n } = useI18n();
+  const tooltip =
```
psiotwo, line 18: This looks strange

image

src/util/Validation.ts
```diff
@@ -0,0 +1,21 @@
...
+const Validation = {
+  QUALITY_AFFECTING_RULES: [
+    "https://slovník.gov.cz/jazyk/obecný/g4",
+    "https://slovník.gov.cz/jazyk/obecný/m1",
+    "https://slovník.gov.cz/jazyk/obecný/g13",
+    "https://slovník.gov.cz/jazyk/obecný/g14",
+  ],
+
+  toSeverity: function (ruleId?: string) {
+    return ruleId && this.QUALITY_AFFECTING_RULES.indexOf(ruleId) !== -1
```
psiotwo, line 15: Cannot see where this manifestates, but I thought that QUALITY_AFFECTING_RULES should end-up with warning, not error.

psiotwo (commented):
see comments

src/component/misc/validation/InputValidationMessage.tsx
ledsoft: Fixed.

src/component/misc/validation/InputValidationMessage.scss
ledsoft: We agreed with @MichalMed that yellow is barely visible. And orange looks too much like error preventing form from being submitted.

src/util/Validation.ts
ledsoft: The severities are
* BLOCKER
* ERROR
* WARNING
Where BLOCKER prevents form from being submitted. ERROR is for violation of rules affecting term quality, the remainder are WARNING. Of course, the enum constants can be adjusted if need be.

psiotwo (approved):
lgtm

PR #177, opened 2021-06-24
1594 remember legend toggle

ledsoft:
https://kbss.felk.cvut.cz/redmine/issues/1594

Annotator remembers legend toggle state.

### DIFF
--- a/src/util/__tests__/BrowserStorage.test.ts
+++ b/src/util/__tests__/BrowserStorage.test.ts
@@ -1,0 +1,42 @@
+import BrowserStorage from "../BrowserStorage";
+
+describe("BrowserStorage", () => {
+
+    const key = "testKey";
+    const value = "testValue";
+
+    beforeEach(() => {
+        localStorage.clear();
+    });
+
+    describe("set", () => {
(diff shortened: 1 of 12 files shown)


src/util/__tests__/BrowserStorage.test.ts
```diff
@@ -0,0 +1,41 @@
+import BrowserStorage from "../BrowserStorage";
+
+describe("BrowserStorage", () => {
```
psiotwo, line 3: Since localStorage is a standard interface I do not see much benefit in having this wrapper object (except of the default value resolution, which can be however also done by initializing the browser storage upfront maybe ...).

psiotwo (approved):
lgtm, although i do not see much value in the wrapper object.

src/util/__tests__/BrowserStorage.test.ts
ledsoft: 1. It simplifies test setup
2. If we decide, for example, to store user's settings on the server or by some other means, they are now behind an interface which can be reimplemented in one place.

PR #179, opened 2021-07-07
[R-#1589] vocabulary name badge to distinguish term context

psiotwo:
- vocabulary/term/document IRIs changed according to the SGoV Scheme in IPR dev/prod
- added vocabulary name badge to exact/related/broader selectors + to the term detail panel - a badge shows up when the term comes from a different vocabulary than the vocabulary of the current term.
- can be tested on https://deploy-preview-179--termit-dev.netlify.app/#/vocabularies/building-act-no183-2006/terms/stavba?namespace=http://onto.fel.cvut.cz/ontologies/slovnik/ and in the respective Edit form.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -11,6 +11,7 @@
         "@formatjs/intl-pluralrules": "^4.0.22",
         "@formatjs/intl-relativetimeformat": "^9.1.2",
         "@fortawesome/fontawesome-free": "^5.15.3",
+        "@opendata-mvcr/assembly-line-shared": "^0.2.0",
         "apexcharts": "^3.26.3",
         "axios": "^0.21.1",
         "bootstrap": "4.6.0",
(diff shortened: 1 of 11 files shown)


MichalMed (commented):
I'm not sure if I get it right, but I do not see the badges in related terms bar, neither in the term detail nor edit view. Tested on listed example and in vocabulary Guyana Africaine.
I also do not get why are the IRIs changed. It does not fit to the methodology used in ODII and KODI.

psiotwo:
> I'm not sure if I get it right, but I do not see the badges in related terms bar, neither in the term detail nor edit view. Tested on listed example and in vocabulary Guyana Africaine.
> I also do not get why are the IRIs changed. It does not fit to the methodology used in ODII and KODI.

On the term detail, you should see the following badge in the broader term list.
image

There is no example yet for related/exact - but you can create your own.

Most of the terms do not have badges, because they are using old IRIs (on dev). All IRIs on IPR-dev and IPR-prod have been changed however - after merge all terms in these deployments should show badge. Changing IRIs in these deployments was needed to allow for using the same IRI scheme as for KODI so that we can use the shared component as well. The idea si to have a single identification scheme and maybe relax/change it centrally.

Would be good to discuss how to align this visualization with the vocabulary visualization in the search results/term detail.

MichalMed (approved):
lgtm

PR #182, opened 2021-07-20
1548 skos broad match

ledsoft:
https://kbss.felk.cvut.cz/redmine/issues/1548

No significant changes, just minor simplification of term creation action.


psiotwo (approved):
lgtm

PR #97, opened 2021-02-23
[R1487] user role change

psiotwo:
- support for changing user roles by admins
- roles are hierarchical, each user should have exactly one role
- changed format of configuration to JSON-LD

PR #193, opened 2021-09-07
Feature/definitionally related terms

ledsoft:
This PR publishes the previously closed source definitionally related terms feature in the public open source version of TermIt.

It also contains implementation of https://kbss.felk.cvut.cz/redmine/issues/1612 frontend. Note that, to verify the functionality, the corresponding backend PR (https://github.com/kbss-cvut/termit/pull/97) must be merged and deployed.


ledsoft:
The corresponding backend is merged and deployed to development, so this PR can be reviewed now.

psiotwo (changes requested):
- the following approval logic does not seem very practical to me. The approval should be possible from the screen where the definition is shown, not in the other direction.
image. I would remove the "Pojmy, v jejichž definici se tento pojem objevil" field from the edit screen (keeping only in the view screen).

- in the view screen - including "Další související pojmy" into "Související pojmy" would make it much more understandable.

- I don't understand what is the logic behind (not) showing the black "(i)" icon. These images are from [this term](https://deploy-preview-193--termit-dev.netlify.app/#/vocabularies/building-act-no183-2006/terms/stavba?namespace=http://onto.fel.cvut.cz/ontologies/slovnik/). Does this mean that in one direction the annotation was done automatically, while in the other it did not (if not how?)? Or does this mean the related term was found in the definition and approved later? If so, both information (automatic analysis origin, approval) should be shown to the user.
image
image

- often, a definition has the form "budovou se rozumí stavba, která ..." - this suggests not a general "related" link between budova and stavba but rather a specialization link. So, finding a term in a definition should be possible to keep "related", or pass to "broader" (or even "exact matches" maybe?).

- a suggestion for another ticket:
concept with prefLabel "stavebni pozemek" and  altLabel "stavba" has been found in the definition of the concept with prefLabel "stavba" - when showing this relationship the user does not understand how is it possible because the altLabel "stavba" is not shown to him in any way. We should either visualize this link (if available), or give some hint to the user.

I didn't try the approval actions yet, neither saw the code.

ledsoft:
PR feedback implemented.

psiotwo (commented):
The edit screen still contains information about the terms in whose definitions the current term appears. It is correct that I cannot edit approve/reject it any more, but even having this information seems useless and confusing on the edit screen.

image

ledsoft:
More PR feedback implementation.

psiotwo (commented):
The edit/view screens look inconsistent:

image

image

What should happen if I remove 'stavební pozemek' from 'related terms'?

ledsoft:
1. Inconsistent in what sense? That there is a "Definition" badge in the view mode and not in the editing mode? There is the definition-related terms edit in the editing mode, that should communicate sufficiently that the terms are related via definition.
2. It is removed from the related terms.

Given the fact that there was basically no description of the acceptance criteria of this feature, it is quite difficult to fulfill them. Perhaps someone else should implement it then...

psiotwo:
> 1. Inconsistent in what sense? That there is a "Definition" badge in the view mode and not in the editing mode? There is the definition-related terms edit in the editing mode, that should communicate sufficiently that the terms are related via definition.
> 2. It is removed from the related terms.
> 
> Given the fact that there was basically no description of the acceptance criteria of this feature, it is quite difficult to fulfill them. Perhaps someone else should implement it then...

My point is only that 'stavební pozemek' can be seen twice on the edit screen (both occurrences are editable), while it is only once on the view screen.

ledsoft:
Well, we are mixing two functionalities (related terms and definitionally-related terms), aren't we. What would be your suggestion?

psiotwo:
> Well, we are mixing two functionalities (related terms and definitionally-related terms), aren't we. What would be your suggestion?

The simplest solution IMO would be to show it twice on the view screen as well (once with badge, once without) - not super-nice, but at least consistent.

I am only afraid that user would be confused if she removes 'stavební pozemek' from related terms and nothing happens on the view screen.

psiotwo (approved):
lgtm

PR #200, opened 2021-10-19
1630 definition related terms simplification

ledsoft:
https://kbss.felk.cvut.cz/redmine/issues/1630

### DIFF
--- a/src/__tests__/environment/Environment.tsx
+++ b/src/__tests__/environment/Environment.tsx
@@ -53,7 +53,7 @@ export function mountWithIntl(
 /**
  * Utility function to flush all pending promises in an async test.
  */
-export function flushPromises() {
+export function flushPromises(): Promise<void> {
   return new Promise((resolve) => {
     scheduler(resolve, 0);
   });
--- a/src/component/MainView.scss
+++ b/src/component/MainView.scss
@@ -123,6 +123,7 @@
 
 .asset-badge {
   width: 4.4rem;
+  vertical-align: text-bottom !important;
 }
 
 /* Reactstrap overrides this by assigning form-control class to all form-controls, including small ones*/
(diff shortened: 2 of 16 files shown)


psiotwo (dismissed):
Cannot see the "definition" badge in the multi-select edit field for related terms.

E.g. https://deploy-preview-200--termit-dev.netlify.app/#/vocabularies/building-act-no183-2006/terms/stavebn%C3%AD-pozemek?namespace=http://onto.fel.cvut.cz/ontologies/slovnik/&includeImported=false

psiotwo (changes requested):
Is the backend up-to-date on dev? 

1. Going to

https://deploy-preview-200--termit-dev.netlify.app/#/vocabularies/building-act-no183-2006/terms/stavebn%C3%AD-pozemek?namespace=http://onto.fel.cvut.cz/ontologies/slovnik/&includeImported=false

I cannot see "definition" badge in the related field of the term edit screen.

2. Having approved "Structure" on the same field, saving it, going back to the edit screen and adding "Structure" once more to the related terms (this time manually, not through the text analysis), removes even the first one from the selector.

ledsoft:
1. Good point, I forgot that we discussed this. I updated the acceptance criteria in the corresponding Redmine ticket and will implement it
2. That looks like a bug, will fix it.

There are no changes to the backend involved (so far, but I don't expect any), so no need to update it.

ledsoft:
Reimplemented according to discussion. Also updated the Redmine task to reflect the discussed acceptance criteria (hopefully).

psiotwo (changes requested):
There seems to be some issue with languages:
On the edit screen the term is rendered as **Structure**, while
on the view screen it is rendered as **Stavba**

ledsoft:
Fixed. However, note that while the definition related terms are using the language facet of a concrete term, the selectors do not and their content is based on the selected UI language. If this is a problem (inconsistent behavior), a separate issue should be created for it.

psiotwo (approved):

PR #224, opened 2022-01-27
1649 vocabulary badge missing

ledsoft:
https://kbss.felk.cvut.cz/redmine/issues/1649

Concerns #218 . Although the exact vocabulary referenced in that issue has a data problem - vocabulary IRI with HTTP instead of HTTPS. A PR was prepared for the library providing vocabulary badge resolution, but was rejected. This PR provides a generic solution to the said issue and should handle vocabularies whose IRIs do not conform to the expected GoV-based pattern.


psiotwo (approved):
lgtm.

Technically OK. 

Although the feature is generally beneficial for arbitrary TermIt deployments, I do not agree, this is the correct solution of the original IPR ticket, it looks more as a hotfix. The correct fix would be to prevent importing vocabularies with non-compliant IRIs.

ledsoft:
1. The current validation of vocabulary IRIs (implemented in [assembly-line-shared](https://github.com/opendata-mvcr/assembly-line-shared)) is tied to Czech legislation (SGoV) vocabularies only. Arbitrary vocabularies (let alone not Czech) would not match at all.
2. Vocabularies created within TermIt would not match this either based on the default configuration we have - https://github.com/kbss-cvut/termit/blob/master/src/main/resources/application.yml

psiotwo:
> 1. The current validation of vocabulary IRIs is tied to Czech legislation (SGoV) vocabularies only. Arbitrary vocabularies (let alone not Czech) would not match at all.
> 2. Vocabularies created within TermIt would not match this either based on the default configuration we have - https://github.com/kbss-cvut/termit/blob/master/src/main/resources/application.yml

Yes, I agree. And I think it is completely consistent with my last message.

ledsoft:
So we are building TermIt for Czech legislation only?

psiotwo:
No and this is why I wrote "Although the feature is generally beneficial for arbitrary TermIt deployments"

ledsoft:
And if we implement the "correct solution" of the original ticket, i.e., prevent importing of vocabularies with IRIs not compatible with the SGoV pattern, aren't we again restricting TermIt (or at least the vocabulary import feature) to Czech legislation?

psiotwo:
> And if we implement the "correct solution" of the original ticket, i.e., prevent importing of vocabularies with IRIs not compatible with the SGoV pattern, aren't we again restricting TermIt (or at least the vocabulary import feature) to Czech legislation?

Well, if we bind valid IRIs to be imported to the IRIs configured in the deploy conf, this should not be a problem.

ledsoft:
But the assembly-line-shared library responsible for resolving the vocabulary badges has the IRI patterns hardcoded.

PR #247, opened 2022-03-16
Task/1651 remove resource management

ledsoft:
https://kbss.felk.cvut.cz/redmine/issues/1651


psiotwo (approved):
Code - TLTR
Functionality w.r.t. the ticket - LGTM. 

A note - having a link from a term to documents that use it, or from a document to confirmed terms might still be valuable.

PR #287, opened 2022-05-18
Bug/214 vocabulary import

ledsoft:
Frontend for #214 fix. See https://github.com/kbss-cvut/termit/pull/152


ledsoft:
@psiotwo has not responded in two weeks, merging.

PR #313, opened 2022-08-22
Feature/165 keycloack

filip-kopecky:
Added keycloak option to the UI. When passed` REACT_APP_AUTHENTICATION=keycloak` in .env the UI is using that auth provider


src/reducer/TermItReducers.ts
```diff
@@ -95,6 +95,8 @@ function messages(state: Message[] = [], action: MessageAction): Message[] {
     case ActionType.LOGOUT:
     case ActionType.LOGIN: // Intentional fall-through
       return [];
+    case ActionType.LOGIN_KEYCLOAK: // Intentional fall-through
```
ledsoft, line 97: There is no fall-through here. Either remove the comment, or modify the code so that the comment corresponds to reality.

filip-kopecky: Sorry about that, I was copying code without much thinking I suppose

src/action/AsyncUserActions.ts
```diff
@@ -42,14 +44,18 @@ export function loadUser() {
...
     return Ajax.get(`${Constants.API_PREFIX}${USERS_ENDPOINT}/current`)
-      .then((data: object) =>
-        JsonLdUtils.compactAndResolveReferences<UserData>(data, USER_CONTEXT)
-      )
+      .then((data: object) => {
+        return JsonLdUtils.compactAndResolveReferences<UserData>(
+          data,
+          USER_CONTEXT
+        );
+      })
```
ledsoft: Unnecessary return results in unnecessary change in formatting.

src/component/login/LoginKeycloak.tsx
```diff
@@ -0,0 +1,47 @@
...
+      </Card>
+    );
+  }
+};
+
+export default connect(undefined, (dispatch: ThunkDispatch) => {
+  return {
+    login: () => dispatch(loginKeycloak),
+  };
+})(injectIntl(withI18n(Login)));
```
ledsoft: Please, rewrite to `useDispatch` and `useI18n` hooks so that we can get rid of the `connect` HOC.

src/component/misc/AuthUnavailable.tsx
```diff
@@ -0,0 +1,16 @@
...
+const AuthUnavailable: React.FC<HasI18n> = ({ i18n }) => {
+  return (
+    <Card>
+      <CardHeader tag="h2">{Constants.APP_NAME}</CardHeader>
+      <CardBody>{i18n("auth.unavailable-message")}</CardBody>
+    </Card>
+  );
+};
+
+export default injectIntl(withI18n(AuthUnavailable));
```
ledsoft: Please, rewrite to the `useI18n` hook and remove the HOCs.

src/reducer/TermItReducers.ts
```diff
@@ -95,6 +95,8 @@ function messages(state: Message[] = [], action: MessageAction): Message[] {
     case ActionType.LOGOUT:
     case ActionType.LOGIN: // Intentional fall-through
       return [];
+    case ActionType.LOGIN_KEYCLOAK:
+      return [];
```
ledsoft: We can probably let it fall-through to the same action, as the result is the same anyway - returning an empty array. So something like
```javascript
case ActionType.LOGOUT:
case ActionType.LOGIN: // Intentional fall-through
case ActionType.LOGIN_KEYCLOAK: // Intentional fall-through
  return [];
```

src/component/MainView.tsx
```diff
@@ -64,6 +65,7 @@ export class MainView extends React.Component<MainViewProps, MainViewState> {
 
   public componentDidMount(): void {
     if (this.props.user === EMPTY_USER) {
+      Routing.saveOriginalTarget();
```
ledsoft, line 70: Could just confirm that this is because Keycloak-based backend returns 302 forward to Keycloak login instead of 401? Because for 401 responses this is handled by Ajax.

ledsoft (changes requested):
See inline comments (but otherwise I tested it and it works).

ledsoft (approved):
LGTM

PR #336, opened 2022-09-30
318 Changes and Comments on snapshots

filip-kopecky:
Added the option to show comments and history of changes on snapshot details. This solution works entirely on the front-end by loading the original history and showing just the part which is relevant to the snapshot. 

@ledsoft I noticed that the vocabulary snapshot summary also includes 'snapshots' (historie revizí). That doesn't make that much sense to me because we are not showing this tab on the snapshotted terms. Should we remove the tab for vocabulary snapshots and add this change to the PR?


ledsoft:
@FilipKopecky Oh yeah, definitely. I must have forgotten to hide the tab. Please, add it to this PR as well. Thanks

filip-kopecky:
Added commit which hides the snapshot history tab. Ready for review @ledsoft

src/component/changetracking/AssetHistory.tsx
```diff
@@ -25,7 +27,28 @@ export const AssetHistory: React.FC<AssetHistoryProps> = (props) => {
   const [records, setRecords] = React.useState<null | ChangeRecord[]>(null);
   React.useEffect(() => {
     if (asset.iri !== Constants.EMPTY_ASSET_IRI) {
-      loadHistory(asset).then((recs) => setRecords(recs));
+      //Check if vocabulary/term is a snapshot
+      if (
+        (asset instanceof Term || asset instanceof Vocabulary) &&
+        asset.snapshotOf()
```
ledsoft, line 33: I thinks `asset.isSnapshot()` would better communicate the intent of the `if` condition

ledsoft (changes requested):
See the inline comment. Otherwise looks good.

src/component/changetracking/AssetHistory.tsx
filip-kopecky: I agree but for some reason it is needed to check whether the `asset.snapshotOf()`has some value. It can happen that the asset returns true for `asset.isSnapshot()` but the property containg information about the original asset is missing, which causes a crash. This would happen everytime a user tried to open a term from a snapshotted vocabulary summary.
I agree that the intent could be expressed more clearly by changing the if statement, but it would require change on the loading part (to ensure that the `asset.snapshotOf()` contains a value) and this was just faster to implement. If you think that the `if statement` is really not clear I can look more into it.
@ledsoft

ledsoft: Ok, I see. Let's leave it as is.

ledsoft (approved):
LGTM

PR #337, opened 2022-10-03
[#318] Snapshot comments fetching via API

filip-kopecky:
Comments for snapshotted terms are now fetched via specific API call. 

Resolves: #318

### DIFF
--- a/src/action/AsyncCommentActions.ts
+++ b/src/action/AsyncCommentActions.ts
@@ -15,13 +15,21 @@ import Comment, {
 } from "../model/Comment";
 import { ErrorData } from "../model/ErrorInfo";
 
-export function loadTermComments(termIri: IRI) {
+export function loadTermComments(
+  termIri: IRI,
+  dateTo?: string,
+  dateFrom?: string
+) {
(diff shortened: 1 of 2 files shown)


src/action/AsyncCommentActions.ts
```diff
@@ -15,13 +15,21 @@ import Comment, {
...
+  dateTo?: string,
+  dateFrom?: string
+) {
   const action = { type: ActionType.LOAD_COMMENTS };
   return (dispatch: ThunkDispatch) => {
     dispatch(asyncActionRequest(action, true));
+    const reqParams: any = {};
+    reqParams.namespace = termIri.namespace;
+    if (dateTo) reqParams.to = dateTo;
+    if (dateFrom) reqParams.from = dateFrom;
```
ledsoft: I think you can actually skip the `if`s, as params with undefined/nullish value will not be added to the request (see `paramsSerializer` in Ajax).

filip-kopecky: You are right, I totally missed that. Added commit which removes the `if`s

ledsoft (approved):
LGTM

PR #350, opened 2022-11-19
[#128] Rename document and files

filip-kopecky:
Added the option to specify the name of the document when creating a vocabulary and also the ability to change it in the edit vocabulary form. Plus added a modal for changing the names of the document's files.

@ledsoft If you don't like the UI, regarding the placement or the behaviour of the components, just let me know and I will fix it right away. 
The backend changes are available in this PR: https://github.com/kbss-cvut/termit/pull/198

### DIFF
--- a/src/component/vocabulary/VocabularyEdit.tsx
+++ b/src/component/vocabulary/VocabularyEdit.tsx
@@ -62,7 +62,7 @@ export class VocabularyEdit extends React.Component<
 
   public onSave = () => {
     const modifiedDocument = Object.assign({}, this.props.vocabulary.document, {
-      label: this.state.documentLabel.trim(),
+      label: this.state.documentLabel?.trim(),
     });
 
     const newVocabulary = new Vocabulary(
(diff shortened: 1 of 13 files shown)


filip-kopecky:
Implemented changes requested by @ledsoft . Entire solution requires only these front-end changes. Back-end PR is closed. 

Resolves: #128

src/component/vocabulary/VocabularySummary.tsx
```diff
@@ -114,6 +118,10 @@ export class VocabularySummary extends EditableComponent<
     });
   };
 
+  public onDocumentSave = (document: Document) => {
```
ledsoft: I would replace the parameter type with `DocumentData`.  Otherwise it is a bit unintuitive why we create a new `Document` from the provided one.

src/component/vocabulary/VocabularyEdit.tsx
```diff
@@ -57,6 +61,10 @@ export class VocabularyEdit extends React.Component<
   };
 
   public onSave = () => {
+    const modifiedDocument = Object.assign({}, this.props.vocabulary.document, {
```
ledsoft, line 64: Could you also add a check for change in the document label so that we do not update the document when there is no change to it?

ledsoft (approved):
LGTM

PR #372, opened 2023-02-03
Enhancement/369 links in attributes

filip-kopecky:
If a value of unmapped property is a valid link, the redirect button appears next to it.
I kept the original output and only added the button, that way the links look the same through the entire application.
@ledsoft The specification did not mention the exact visual style, so if you have any remarks regarding the look, I am ready to change it to your liking.

Closes: #369


ledsoft (approved):
LGTM

Issue #293, opened 2022-06-15
Consider replacing react-breadcrumbs


ledsoft:
[React-breadcrumbs](https://github.com/robbestad/react-breadcrumbs) is no longer developed and it has a peer dependency on React 16, which causes `npm install` issues since we are already on React 17.

We should consider replacing the library with something more up-to-date with equivalent functionality.

filip-kopecky:
There are two options how to handle this issue.

React Router in v6.4 supports react-breadcrumbs natively. I think it would be better for longevity to update React router to the latest version and use the breadcrumbs described [here](https://reactrouter.com/en/main/hooks/use-matches#breadcrumbs).

Other option is to stay on current version of React Router and use this [npm package](https://github.com/icd2k3/react-router-breadcrumbs-hoc). It would definetely be faster to implement, but it is a temporary fix.

Question is, how important this issue is. If the migration to React Router is something which would have to be done sooner or later, this might be a good opportunity to do so.  @ledsoft ?

ledsoft:
Ok, I guess it makes sense to bite the bullet and do the migration now, because we would have to do it eventually. So please, go ahead and implement them using React Router v6.

PR #374, opened 2023-02-09
Enhancement/293 custom breadcrumbs

filip-kopecky:
Removed react-breadcrumbs package.
Took the package code and rewrote it into functional components. 
The interface of the components stayed the same.

Closes: #293


ledsoft (approved):
LGTM, nice work!

PR #385, opened 2023-02-22
Fix/384 missing import terms

filip-kopecky:
Fixes the missing terms in the term list. It was caused by `processTermsForTreeSelect` which filters out terms that are not from passed array of vocabulary IRIs. However the passed array contained more information than only IRIs (e.g. context). These additional information are not needed for the filter to work.

I am not entirely sure about the test, I rewrote in somewhat not pleasing way. If there is a generator or some custom function that would make it more pleasing to look at, please @ledsoft let me know. But this approach works just fine while having minimal amount of changes applied to the test. 

Resolves: #384

### DIFF
--- a/src/component/public/term/Terms.tsx
+++ b/src/component/public/term/Terms.tsx
@@ -107,9 +107,9 @@ export class Terms extends React.Component<GlossaryTermsProps, TermsState> {
       )
       .then((terms) => {
         const matchingVocabularies = this.state.includeImported
-          ? Utils.sanitizeArray(
-              this.props.vocabulary!.allImportedVocabularies
-            ).concat(this.props.vocabulary!.iri)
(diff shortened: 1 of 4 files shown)


ledsoft:
@filip-kopecky , thanks a lot for looking into it. However, the PR does not actually resolve the issue. The purpose of `allImportedVocabularies` is to include transitively imported vocabularies, which is what `importedVocabularies` does not (it contains only directly imported ones). So the PR fix would not work in case a vocabulary transitively imports other vocabularies.
I looked at the contents of the `allImportedVocabularies` and, as you mentioned, it contains more data. In fact, the server response is not a valid JSON, so this is a server-side issue (more specifically a regression in one of our libraries). I will fix it.
I am closing this PR as it would not resolve the issue fully, but thanks again for finding the bug and looking into it.

Issue #235, opened 2022-02-14
Badges shall show vocabulary name and redirect to it


MichalMed:
Hover the mouse over badge of a vocabulary term shall show the full name (or IRI) of a vocabulary, clicking on a badge shall redirect to the vocabulary detail.

lindatothova:
Nice to have, not a priority now :-)

PR #388, opened 2023-02-26
Enhancement/235 badges upgrade

filip-kopecky:
### !!! Please read this before looking at the code !!!
I added new vocabulary badge that can be clicked and users are then redirected accordingly to the vocabulary summary.
I only added this clickable badge to the `Term summary`. I think it would be impractical to also have it in the `InteligentTree renders`. Users could click the badge by accident and lose all their progress while for example modyfing term .

Regarding the implementation there are couple of notes I need to say. The tooltip implementation is usually done by creating a unique ID and assigning it to the targeted element. That works well until tests come into the picture. All of them fail because they cannot access DOM. So I used refs instead and conditional rendering. Another issue was inconsistent look of the tooltip content. When using `AssetLabel` for the term label, the tooltip looked out of place. It is because once the tooltip is generated, the dimensions of the content are set until it is closed again. `AssetLabel` at first returns IRI and then the label. That caused mismatch between the original `IRI` size and the `label` size of the tooltip.

@ledsoft I know that the code might look messy, but I really did not have any other choice. If you disagree with anything, please let me know and I will fix it immediately. 

Resolves: #235


filip-kopecky:
After discussion with @ledsoft we agreed to prefer custom ids over refs. Now the component requires a `section` property to actually identify where the badge is located + iri of the term. It can happen that a term is present in `exactMatches` category, as well as in  the `relatedTerms` category. If we wouldn't provide the context where the badge is located, the component would not know where the tooltip should be shown.

About the tests. I mocked the tooltip in the problematic test. The mocking of popper doesn't have any effect on the test. I guess we could rewrite the component to use `Popover` but that seems like unnecessary overkill

src/component/vocabulary/VocabularyNameBadgeButton.tsx
```diff
@@ -0,0 +1,78 @@
...
+  const iri = VocabularyUtils.create(vocabulary!.iri!);
+  const path = Routing.getTransitionPath(
+    isLoggedIn(user)
+      ? Routes.vocabularySummary
+      : Routes.publicVocabularySummary,
+    {
+      params: new Map([["name", iri.fragment]]),
+      query: new Map([["namespace", iri.namespace!]]),
+    }
+  );
```
ledsoft, line 60: Could use `Vocabularies.getVocabularyRoutingOptions` from Routing.

src/component/vocabulary/VocabularyNameBadgeButton.tsx
```diff
@@ -0,0 +1,78 @@
...
+import { Link } from "react-router-dom";
+import { ThunkDispatch } from "../../util/Types";
+import { getLabel } from "../../action/AsyncActions";
+import Utils from "../../util/Utils";
+
+interface VocabularyNameBadgeButtonProps {
+  vocabulary?: AssetData;
+  className?: string;
+  termIri: string;
+  section: string;
```
ledsoft: I would make it optional with empty string as default.

src/component/vocabulary/VocabularyNameBadgeButton.tsx
filip-kopecky: I looked into the method and it only accepts `Vocabulary` as a parameter. The components which render the badge have the vocabulary only as an `AssetData`. Is there a way how to convert it? I couldn't find any... The code is not that long, so I don't think it is doing that much harm, but I get your point. 
So do we leave it? Or should I fetch the vocabularies? Or is there a way how to sufficiently parse `AssetData` into `Vocabulary`? @ledsoft

ledsoft: Ok, sorry, I saw the variable `vocabulary` and didn't bother checking its type. Let's leave it as is.

ledsoft (approved):
LGTM

PR #483, opened 2024-07-23
[Enhancement #482] Support sending invitation links to new users

### LINKED ISSUE #482: Support sending invitation links to new users
# Description
As an admin, I want TermIt to be able to send invitation link that a new user will be able to use to access TermIt and set their password, so that I don't have to set the password and them give it to them by (possibly) unsecure means.

However, I want to be able to set the password if necessary. So it should be possible to choose whether to send an invitation link or set the password.

# Usage
Inviting new users without setting password for them.
****

lukaskabc:
Resolves #482

### DIFF
--- a/src/action/AsyncUserActions.ts
+++ b/src/action/AsyncUserActions.ts
@@ -213,8 +213,8 @@ export function createNewUser(user: UserAccountData) {
   const action = { type: ActionType.CREATE_USER };
   return (dispatch: ThunkDispatch) => {
     dispatch(asyncActionRequest(action));
-    return Ajax.put(
-      Constants.API_PREFIX + "/users",
+    return Ajax.post(
+      Constants.API_PREFIX + "/admin/users",
       content(user).contentType("application/json")
     )
       .then(() => dispatch(asyncActionSuccess(action)))
@@ -469,9 +469,10 @@ export function doesUsernameExists(username: string) {
       `${Constants.API_PREFIX}${USERS_ENDPOINT}/username`,
       params({ username })
     )
-      .then((data) =>
-        dispatch(asyncActionSuccessWithPayload(action, data === true))
-      )
+      .then((data) => {
+        dispatch(asyncActionSuccess(action));
+        return data === true;
+      })
       .catch((error: ErrorData) => dispatch(asyncActionFailure(action, error)));
   };
 }
--- a/src/action/AsyncUserActions.ts
+++ b/src/action/AsyncUserActions.ts
@@ -460,3 +460,18 @@ export function loadManagedAssets(user: User) {
       });
   };
 }
+
+export function doesUsernameExists(username: string) {
+  const action = { type: ActionType.DOES_USERNAME_EXISTS };
+  return (dispatch: ThunkDispatch) => {
+    dispatch(asyncActionRequest(action, true));
+    return Ajax.get(
+      `${Constants.API_PREFIX}${USERS_ENDPOINT}/username`,
+      params({ username })
+    )
+      .then((data) =>
+        dispatch(asyncActionSuccessWithPayload(action, data === true))
+      )
+      .catch((error: ErrorData) => dispatch(asyncActionFailure(action, error)));
+  };
+}
--- a/src/action/AsyncUserActions.ts
+++ b/src/action/AsyncUserActions.ts
@@ -213,7 +213,7 @@ export function createNewUser(user: UserAccountData) {
   const action = { type: ActionType.CREATE_USER };
   return (dispatch: ThunkDispatch) => {
     dispatch(asyncActionRequest(action));
-    return Ajax.post(
+    return Ajax.put(
       Constants.API_PREFIX + "/users",
       content(user).contentType("application/json")
     )
--- a/src/component/administration/user/CreateNewUserForm.tsx
+++ b/src/component/administration/user/CreateNewUserForm.tsx
@@ -47,7 +47,6 @@ const CreateNewUserForm: React.FC<CreateNewUserFormProps> = (props) => {
   const fetchUsernameExists = useRef(
     _.debounce((username) => {
       dispatch(doesUsernameExists(username)).then((action) => {
-        console.debug(username, action);
         if (action.status === AsyncActionStatus.SUCCESS) {
           const actionSuccess = action as AsyncActionSuccess<boolean>;
           setUsernameExists(actionSuccess.payload);
--- a/src/component/administration/user/CreateNewUserForm.tsx
+++ b/src/component/administration/user/CreateNewUserForm.tsx
@@ -19,11 +19,11 @@ import {
   Row,
   UncontrolledTooltip,
 } from "reactstrap";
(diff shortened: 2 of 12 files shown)


src/action/AsyncUserActions.ts
```diff
@@ -460,3 +460,18 @@ export function loadManagedAssets(user: User) {
...
+export function doesUsernameExists(username: string) {
+  const action = { type: ActionType.DOES_USERNAME_EXISTS };
+  return (dispatch: ThunkDispatch) => {
+    dispatch(asyncActionRequest(action, true));
+    return Ajax.get(
+      `${Constants.API_PREFIX}${USERS_ENDPOINT}/username`,
+      params({ username })
+    )
+      .then((data) =>
+        dispatch(asyncActionSuccessWithPayload(action, data === true))
```
ledsoft: If you do not have a reducer for this action (and I don't see it anywhere), there is no point in calling `asyncActionSuccessWithPayload` and you can use just `asyncActionSuccess` and return the result.

src/action/AsyncUserActions.ts
```diff
@@ -213,7 +213,7 @@ export function createNewUser(user: UserAccountData) {
   const action = { type: ActionType.CREATE_USER };
   return (dispatch: ThunkDispatch) => {
     dispatch(asyncActionRequest(action));
-    return Ajax.post(
+    return Ajax.put(
```
ledsoft: See my comment on the corresponding backend PR - I don't understand why you changed the HTTP method from POST (which is supposed to be used for creating resources) to PUT (which is supposed to be used for updating resources).

src/component/administration/user/CreateNewUserForm.tsx
```diff
@@ -0,0 +1,285 @@
...
+  // ref is required to keep the same debounced function between renders
+  const fetchUsernameExists = useRef(
+    _.debounce((username) => {
+      dispatch(doesUsernameExists(username)).then((action) => {
+        if (action.status === AsyncActionStatus.SUCCESS) {
+          const actionSuccess = action as AsyncActionSuccess<boolean>;
+          setUsernameExists(actionSuccess.payload);
+        }
+      });
+    }, 250)
```
ledsoft: What is this magical constant? What does it do? If it is necessary here, at least explain its use and value in a comment.

src/component/administration/user/CreateNewUserForm.tsx
```diff
@@ -0,0 +1,285 @@
...
+    }
+    // if there is at least one "false" value, the form is invalid
+    return !validationResults
+      // map validation results to booleans
+      .map((result) => result.severity === Severity.VALID)
+      // search "false" value
+      .includes(false);
+  };
+
+  const renderPasswordInputs = () => {
```
ledsoft: Why is this a function when it is called only once anyway?

src/component/administration/user/CreateNewUserForm.tsx
```diff
@@ -0,0 +1,285 @@
...
+    return validateNonEmpty(firstName);
+  };
+
+  const validateLastName = () => {
+    return validateNonEmpty(lastName);
+  };
+
+  const validatePassword = () => {
+    return validateNonEmpty(password);
+  };
```
ledsoft: I would inline these three functions, they don't provide much more info and the component is already way too long (imagine writing a function with 280 LOC in Java)

src/i18n/cs.ts
```diff
@@ -95,13 +95,26 @@ const cs = {
...
     "resetPassword.submit": "Změnit heslo",
     "resetPassword.success": "Heslo změněno",
     "resetPassword.invalidToken":
-      "Neplatný nebo expirovaný link pro změnu hesla. Zažádejte o změnu hesla znovu.",
+      "Neplatný nebo expirovaný odkaz pro změnu hesla. Zažádejte o změnu hesla znovu.",
+
+    "createPassword.passwordsNotEqual": "Heslo a jeho potvrzení se neshodují.",
+    "createPassword.mask": "Vytvářím heslo...",
+    "createPassword.title": "Vytvoření hesla",
+    "createPassword.subtitle": "Vytvoření hesla",
```
ledsoft: What's the point of two constants with the same value?

ledsoft (commented):
See the inline comments. Mostly minor stuff, the main issue I have is the change of the HTTP method for creating an account.

src/action/AsyncUserActions.ts
lukaskabc: What about [result processing](https://github.com/kbss-cvut/termit-ui/blob/28f1b32e6877779648a1cb01507391f747204ff5/src/component/administration/user/CreateNewUserForm.tsx#L50)

I think it's better to have the possible promise result types "close together"  
`Promise<AsyncActionSuccess<boolean> | AsyncFailureAction>` looks more consistent than  
`Promise<boolean | AsyncFailureAction>`.
I can check for the actual value of `AsyncAction.status` with the former, and with the latter, I would check for the type `typeof result == "boolean"`.

But maybe it's just personal preference. I can rewrite it.

ledsoft: I don't like leaking async handling into components, that's the main reason. But I understand your reasoning w.r.t. close return types. I suppose the generic error handling in async actions is not as beneficial as I hoped it would be. Let's keep it this way then.

src/component/administration/user/CreateNewUserForm.tsx
```diff
@@ -0,0 +1,272 @@
...
+      .includes(false);
+  };
+
+  return (
+    <Container>
+      <PromiseTrackingMask area="CreateNewUserForm" />
+      <Form>
+        <Row>
+          <Col md={6}>
+            <EnhancedInput
```
ledsoft: The inputs should be all small

src/i18n/cs.ts
```diff
@@ -204,6 +214,14 @@ const cs = {
     "administration.users.types.admin": "Tento uživatel je administrátor",
     "administration.users.oidc":
       "Pro správu uživatelů je využívána externí autentizační služba.",
+    "administration.users.create.password-toggle.user":
+      "Nechat nového uživatele vytvořit heslo",
```
ledsoft: "Poslat aktivační email"

src/i18n/en.ts
```diff
@@ -194,6 +204,14 @@ const en = {
     "administration.users.types.admin": "This user is an administrator",
     "administration.users.oidc":
       "An external authentication service is used to manage users.",
+    "administration.users.create.password-toggle.user":
+      "Have the new user create a password",
```
ledsoft: "Send activation email"

ledsoft (commented):
One more round of comments after actually testing the functionality.

src/component/administration/user/CreateNewUserForm.tsx
lukaskabc: every input with `md={6}` ?

ledsoft: Every input in the app. Can be probably handled by passing it in `EnhancedInput` which, in contrast to `CustomInput` does not do it.

PR #493, opened 2024-07-30
[Bug #491] Navigation between files from different vocabularies causes term selector to show cached terms

### LINKED ISSUE #491: Navigation between files from different vocabularies causes term selector to show cached terms
# Description
If I open a file in one vocabulary, then go to another vocabulary and open a file there, term selector offers me terms from both vocabularies.

The problem is in twofold:
1. Reducer does not clear `annotatorTerms` on fetch of new ones
2. `ResourceFileDetail` uses stale redux state (`state.selectedFile`) to resolve text analysis records, resource detail and other critical data.

# Severity
It is not a critical problem because annotator is able to load relevant terms eventually, but it does so by loading terms one by one, which is highly inefficient

lukaskabc:
Bug #491 
The component uses the redux state that is kept between component mounts instead of the inner component state.
When the second component mount is made, data (vocabulary IRI) from the previous life cycle is still available.

An outgoing request for new data will eventually replace the data from the previous life cycle. 
Still, the old data (vocabulary IRI) is propagated to the child components, which will request to load additional data (loadTerms).   When the old data is replaced by the new data (new vocabulary IRI), there may still be a pending request from a child component to load the (now outdated) data. 
Because of this, the new request with the new parameters is being effectively canceled.

This PR introduces a way to include AbortController in AsyncAction, which allows the cancellation of the outgoing request when a new one is made with recent data.

An alternative to this solution would probably be rewriting related components to use their inner state instead of the redux state (use the `dispatch(/* action */).then(data => setData(data))` syntax).

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3711,7 +3711,7 @@
       "version": "17.0.20",
       "resolved": "https://registry.npmjs.org/@types/react-dom/-/react-dom-17.0.20.tgz",
       "integrity": "sha512-4pzIjSxDueZZ90F52mU3aPoogkHIoSIDG+oQ+wQK7Cy2B9S+MvOqY0uEA/qawKz381qrEDkvpwyt8Bm31I8sbA==",
-      "devOptional": true,
+      "dev": true,
       "dependencies": {
         "@types/react": "^17"
       }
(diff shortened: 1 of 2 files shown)


ledsoft:
While this seems like an interesting solution, I feel it is overly complicated. I thought the simplest solution would be to evict the `selectedFile` state value whenever a (different) vocabulary is loaded. Or it could actually be evicted whenever the annotator is closed, as we don't have any other dedicated file detail view anyway.
This would spare us the aborted request completely.
That is not to say I don't think action cancellation is not a useful feature, I just don't think it is necessary in this particular case.
@lukaskabc Can you check whether the solution I proposed is enough to fix the problem? If so, split this PR into two - one simple to fix the linked issue and another with added support for async action cancellation (I am sure we will find it useful at some point).

lukaskabc:
@ledsoft the proposed solution is enough to fix the problem

ledsoft (approved):
LGTM

PR #494, opened 2024-07-31
Add option to abort pending AsyncAction request

lukaskabc:
created on the basis of PR https://github.com/kbss-cvut/termit-ui/pull/493

### DIFF
--- a/src/reducer/TermItReducers.ts
+++ b/src/reducer/TermItReducers.ts
@@ -441,7 +441,7 @@ function notifications(
 }
 
 function pendingActions(
-  state: { [key: string]: AsyncActionStatus } = {},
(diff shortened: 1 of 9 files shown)


src/reducer/TermItReducers.ts
```diff
@@ -441,7 +441,7 @@ function notifications(
 }
 
 function pendingActions(
-  state: { [key: string]: AsyncActionStatus } = {},
+  state: { [key: string]: AsyncActionStatus | AbortController } = {},
```
ledsoft: I am not a fan of using two completely different types of objects as values here. Wouldn't it be in this case better to introduce a new type that would contain both the action status and the optional abort controller?

ledsoft (approved):
LGTM

PR #499, opened 2024-08-07
[Enhancement kbss-cvut/termit-ui#473] Annotator: resolve "memory leak" and reset filter on occurrence marking

lukaskabc:
resolves kbss-cvut/termit-ui#473

fixes updating the state of an unmounted element

### DIFF
--- a/src/action/AsyncAnnotatorActions.ts
+++ b/src/action/AsyncAnnotatorActions.ts
@@ -58,30 +58,52 @@ export function loadAllTerms(
 }
 
 // Cache of pending term fetches, used to prevent repeated concurrent attempts at fetching the same term
-const pendingTermFetches: { [key: string]: Promise<Term | null> } = {};
+const pendingTermFetches: {
+  [key: string]: {
+    promise: Promise<Term | null>;
+    abortController: AbortController;
+  };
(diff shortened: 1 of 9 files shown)


src/action/AsyncAnnotatorActions.ts
```diff
@@ -58,30 +58,52 @@ export function loadAllTerms(
...
+} = {};
 
-export function loadTermByIri(termIri: string) {
+export function loadTermByIri(
+  termIri: string,
+  abortController: AbortController = new AbortController()
+) {
   const action = { type: ActionType.ANNOTATOR_LOAD_TERM };
   return (dispatch: ThunkDispatch, getState: GetStoreState) => {
+    // TODO: remove throwIfAborted null check (problem with NodeJS version) support from 17.3.0 (Released 2021-12-17)
```
ledsoft, line 74: Let's bump minimum required Node version to 18. I recently resolved an issue with building TermIt on 20 and later, so there is no reason to stick with such an old version. Browser support is also good (since start of 2022).

ledsoft (approved):
LGTM, I created a follow-up ticket for the Node upgrade.

PR #512, opened 2024-09-11
Loading text analysis from WS and displaying long-running tasks

lukaskabc:
see kbss-cvut/termit#295

image


src/component/main/LongRunningTasksStatus.tsx
```diff
@@ -0,0 +1,155 @@
...
+      tasksToUpdate.delete(task);
+    };
+  }, [tZero, tasksToUpdate]);
+
+  const date = new Date(time);
+  const hours = date.getUTCHours() > 0 ? date.getUTCHours() + ":" : "";
+  const minutes = padStart(date.getUTCMinutes().toString(), 2, "0");
+  const seconds = padStart(date.getUTCSeconds().toString(), 2, "0");
+
+  return <>{`${hours}${minutes}:${seconds}`}</>;
```
ledsoft, line 154: What about using TimeAgo to display the duration of the task? We already use for displaying time since last edit/comment.

src/component/term/TermDetail.tsx
```diff
@@ -205,9 +223,12 @@ export class TermDetail extends EditableComponent<
...
     if (!this.state.edit) {
       actions.push(
-        <IfNotInTerminalState term={this.props.term!}>
+        <IfNotInTerminalState
+          term={this.props.term!}
+          key="term.detail.edit.if.not.terminal"
+        >
           <IfVocabularyActionAuthorized
-            key="term-detail-edit"
+            key="term.detail.edit.if.authorized"
```
ledsoft: Why change the key? And the change is not consistent, the remove action below has key `term.detail.remove.if` without the `.authorized` at the end.

src/component/vocabulary/VocabularySummary.tsx
ledsoft, line 1: Why the key changes? Makes the PR larger than necessary? If we want to start using dot-notation for keys, I propose to do it in a separate PR for the whole application.

src/i18n/cs.ts
```diff
@@ -870,6 +870,21 @@ const cs = {
     "auth.action.unauthorized": "K této akci nemáte dostatečná práva.",
     "auth.view.unauthorized": "K zobrazení nemáte dostatečná práva.",
     "auth.notEditable.message.readOnly": "Tento { type } je pouze pro čtení.",
+
+    "longrunningtasks.description": "V systému právě probíhá:",
+    "longrunningtasks.state.done": "Dokončeno",
+    "longrunningtasks.state.pending": "Bude spuštěno",
+    "longrunningtasks.state.running": "Probíhá",
+    "longrunningtasks.name.termDefinitionAnalysis":
+      "Textová analýza definice pojmu",
```
ledsoft: I would say just 'analýza definice pojmu', without specifying that it is text analysis (holds for all the new constants here).

src/i18n/en.ts
```diff
@@ -859,6 +859,21 @@ const en = {
     "auth.action.unauthorized": "You have insufficient rights for this action.",
     "auth.view.unauthorized": "You have insufficient rights to view this.",
     "auth.notEditable.message.readOnly": "This { type } is read only.",
+
+    "longrunningtasks.description": "The system is currently running:",
+    "longrunningtasks.state.done": "Completed",
+    "longrunningtasks.state.pending": "Pending",
+    "longrunningtasks.state.running": "Running",
+    "longrunningtasks.name.termDefinitionAnalysis":
+      "Text analysis of a term definition",
```
ledsoft: Same as for the Czech localization - remove the 'text' and just say it is 'analysis'

src/reducer/WebSocketUtilityDispatchers.ts
```diff
@@ -0,0 +1,33 @@
...
+    }
+
+    const tasks: { [uuid: string]: LongRunningTask } = JSON.parse(message.body);
+
+    const mapped = Object.keys(tasks).map((uuid) => {
+      const task = tasks[uuid];
+      if (task.startedAt) {
+        // @ts-ignore
+        task.startedAt = new Date(task.startedAt * 1000);
+        task.startedAt.setMilliseconds(0); // round to second
```
ledsoft: If the `startedAt` value comes from the server, I propose to do the rounding on server. Java Time API has much nicer API for this and it would be more consistent from the clients' point of view.

ledsoft: And if it doesn't come from the server, the rounding should be done where the `startedAt` value is initialized.

ledsoft (changes requested):
A few minor comments, otherwise looks good (not tested, yet).

src/component/vocabulary/VocabularySummary.tsx
lukaskabc: I was looking for a missing key and wanted to make them more consistent as they somewhere use hyphens and elsewhere dots; changes reverted while keeping the new key

src/component/main/LongRunningTasksStatus.tsx
lukaskabc: I wanted to show how long the task has been running.

src/reducer/WebSocketUtilityDispatchers.ts
lukaskabc: truncated on the backend [LongRunningTaskStatus.java#L25](https://github.com/kbss-cvut/termit/blob/34d398ea1514ccd6d97acc7876d402f3a4c30c37/src/main/java/cz/cvut/kbss/termit/util/longrunning/LongRunningTaskStatus.java#L25)

ledsoft (approved):
LGTM, nice work!

PR #514, opened 2024-09-13
[Enhancement #505] Support for sending parameterized error messages to client

### LINKED ISSUE #505: Supporting sending parameterized error messages to client
Currently, the backend returns error objects that contain an error message in English and an identifier of a message that can be localized by the client. However, some errors contain dynamic data such as term label/identifier which is lost in the localized message because there is no way to send them to client at the moment.

Error messages sent from the server should support these values. It could work the same way react-intl `formatMessage` works, where there is the message id and then a generic object with key-value pairs that can be used to populate the message.

lukaskabc:
resolves #505

### DIFF
--- a/src/component/audit/ErrorLogViewer.tsx
+++ b/src/component/audit/ErrorLogViewer.tsx
@@ -43,15 +43,9 @@ const ErrorLogViewer: React.FC = () => {
           {errors.map((item) => {
             let error = item.error;
             if (error.messageId) {
-              if (error.values && Object.keys(error.values).length > 0) {
-                error = Object.assign({}, error, {
-                  message: formatMessage(error.messageId, error.values),
-                });
-              } else {
-                error = Object.assign({}, error, {
-                  message: i18n(error.messageId),
-                });
-              }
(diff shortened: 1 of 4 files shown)


src/component/audit/ErrorLogViewer.tsx
```diff
@@ -43,9 +43,15 @@ const ErrorLogViewer: React.FC = () => {
           {errors.map((item) => {
             let error = item.error;
             if (error.messageId) {
-              error = Object.assign({}, error, {
-                message: i18n(error.messageId),
-              });
+              if (error.values && Object.keys(error.values).length > 0) {
+                error = Object.assign({}, error, {
+                  message: formatMessage(error.messageId, error.values),
```
ledsoft: What if you used `formatMessage` even when error.values is empty? Wouldn't it behave the same way as `i18n`?

lukaskabc: Yes, it behaves the same way.

ledsoft (approved):
LGTM

PR #518, opened 2024-09-19
[Bug #511] Add translations for invalid identifier errors

### LINKED ISSUE #511: TermIt does not prevent manual setting of invalid URI as identifier
# Problem description

While TermIt identifier generator generates only valid URIs, because the user is able to manually edit the identifier, they are able to set an invalid URI as identifier value. In this case, TermIt should refuse the identifier with a reasonable error message so that the user is able to fix the problem.

**Example:**
`http://agentura-cas.cz/DSS---vlastnosti/class/delka-dostriku-[m]`

# Solution alternatives

1. Validate the identifier input immediately in the UI and prevent saving the asset (Term, Vocabulary) when the identifier is not a valid URI
2. Validate the identifier on the backend (which will probably be necessary anyway, because a REST client might send data with invalid identifier, bypassing the UI) and return a reasonable error message.

lukaskabc:
resolves #511 

adds translation messages for kbss-cvut/termit#300

### DIFF
--- a/src/i18n/cs.ts
+++ b/src/i18n/cs.ts
@@ -799,6 +799,10 @@ const cs = {
     "error.vocabulary.import.excel.labelWithDifferentIdentifierExists":
       'Slovník již obsahuje pojem s názvem "{label}" s rozdílným identifikátorem než s jakým byl importován. Existující identifikátor: {existingUri}',
 
+    "error.invalidUriCharacter":
+      'Neplatný identifikátor: "{uri}", neočekával se znak "{char}" na pozici {index}.',
+    "error.invalidIdentifier": 'Neplatný identifikátor: "{uri}"',
+
     "history.label": "Historie změn",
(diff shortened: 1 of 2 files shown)


src/i18n/cs.ts
```diff
@@ -799,6 +799,10 @@ const cs = {
     "error.vocabulary.import.excel.labelWithDifferentIdentifierExists":
       'Slovník již obsahuje pojem s názvem "{label}" s rozdílným identifikátorem než s jakým byl importován. Existující identifikátor: {existingUri}',
 
+    "error.invalidUriCharacter":
+      'Neplatný identifikátor: "{uri}", neočekával se znak "{char}" na pozici {index}.',
```
ledsoft: "neočekávaný znak"

ledsoft (approved):
LGTM

PR #592, opened 2024-12-10
[Enhancement #585] WebSocket connect improvements

### LINKED ISSUE #585: WebSocket connect improvements
When the WebSocket is unable to connect (typically when the user's JWT is expired), it attempts to connect repeatedly even though it will get the same error. Instead, it should cease any further attempts until the user logs in.

Another issue seems to be that even after successful login the socket is unable to connect (at least when I was testing it locally).

lukaskabc:
When the WebSocket receives an error, it will mark the current JWT token as invalid and not attempt another connection until the token is refreshed.

Resolves #585

### DIFF
--- a/src/WebSocketApp.tsx
+++ b/src/WebSocketApp.tsx
@@ -96,7 +96,7 @@ export const WebSocketWrapper: React.FC<{
 
   useEffect(() => {
     const callback = () => {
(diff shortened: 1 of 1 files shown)


ledsoft (approved):
LGTM

PR #593, opened 2024-12-10
[Enhancement #535] Ability to clear validation/text analysis queue

### LINKED ISSUE #535: Ability to clear validation/text analysis queue
# Description
As an administrator, I want to be able to clear the validation/text analysis queue. It should be possible to do it from the administration page. However, this would not cancel already running tasks, only those scheduled.

# Usage
Sometimes there may be need to clean up the queue in case there is an error in task processing. Instead of having to restart the system, it would be nice to be able to do so from the administration UI.

****

lukaskabc:
Adds button to the maintenance-administration for clearing the long-running tasks queue (background processes).


ledsoft (approved):
LGTM

PR #603, opened 2025-01-29
[Enhancement #481] Allow disabling public view of all vocabularies/ vocabulary

### LINKED ISSUE #481: Allow disabling public view of all vocabularies/ vocabulary
# Description
As a vocabulary maintainer, I want to be able to either prevent the vocabulary from showing in the public view (without preventing users in role "READER" from seeing it, as it is currently implemented) or disable the public view altogether.

# Usage
If the vocabulary I work on is not to be shared outside the application, yet I want to be able to make it read-only for users.

# Usefulness of the Feature

1. The scenario cannot be realized without it
****

lukaskabc:
Updates the UI with a new anonymous user role.

Restricts that 
- anonymous user role can only have access level <write
- restricted user role can only have access level <security
- a user group can only have access level <security


ledsoft (approved):
LGTM

Issue #608, opened 2025-03-04
Improve efficiency of TermIt UI


ledsoft:
TermIt UI often makes duplicite requests to the backend or loads more data than necessary. This introduces unnecessary load on the backend and creates excessive network traffic. We should try to reduce this as much as possible.

PR #612, opened 2025-03-06
Perf#608 improve efficiency

ledsoft:
Introduces term info cache to decrease traffic between frontend and backend, mainly on the Vocabulary detail view (Activity tab).

Part of #608.


ledsoft:
@lukaskabc Just FYI, no need to take any actions, but feel free to provide feedback if interested.

PR #652, opened 2025-07-31
[Fix] Fix Term definition source link using document instead of vocabulary causing invalid linking to annotator

lukaskabc:
Button in term detail for navigation to term definition source was redirecting to annotator using a document IRI, resulting in an invalid URL `/vocabularies/document/document/<file>`. The issue is resolved by using the vocabulary IRI if it is available, resulting in a correct URL  `/vocabularies/<vocabulary>/document/<file>`.


ledsoft (approved):
LGTM, nice catch!

PR #660, opened 2025-08-18
[Enhancement #659] Restrict full-text search language selection to available languages

### LINKED ISSUE #659: Restrict full-text search language selection to available languages
The full-text search language selector (#430) currently allows selecting any of the languages supported by the ISO package we use. However, in reality only a small subset of languages is actually used in the data. It would make sense to restrict the selection only to languages actually present in the data.

lukaskabc:
resolves #659
- Loads indexed languages from the configuration endpoint
- Restricts the full-text search language selection to indexed languages
- Allows all language options in a case when no indexed language is found
- Indexed languages are filtered only to valid options according to Intl

### DIFF
--- a/src/component/search/label/NavbarSearch.tsx
+++ b/src/component/search/label/NavbarSearch.tsx
@@ -234,7 +234,9 @@ export class NavbarSearch extends React.Component<
             onChange={this.onChange}
             onKeyPress={this.onKeyPress}
           />
-          {languageSelect}
+          {!!this.props.indexedLanguages?.length &&
+            this.props.indexedLanguages.length > 1 &&
+            languageSelect}
           {!navbar && searchIcon}
           {this.props.searchString && clearIcon}
         </InputGroup>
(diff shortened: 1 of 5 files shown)


ledsoft (changes requested):
After reconsidering it, I still think it is better to not show the language selector and default to no language for FTS when backend provides no indexed languages (typically meaning there were no texts to index on the last run of the initializer). I think it is also more consistent with the idea of the ticket...

lukaskabc:
Okey, in that case, should the selector be also hidden if there is only one language option?

ledsoft:
Good point, I think yes. If there is nothing to select, there is no point showing the selector (I read this in some UX article on Nielsen I think).

src/component/search/label/NavbarSearch.tsx
```diff
@@ -220,7 +234,9 @@ export class NavbarSearch extends React.Component<
             onChange={this.onChange}
             onKeyPress={this.onKeyPress}
           />
-          {languageSelect}
+          {!!this.props.indexedLanguages?.length &&
```
ledsoft: Can we just do `Utils.sanitizeArray(this.props.indexedLanguages) > 1`? I am not a fan of such complex rendering conditions, when they can be avoided.

lukaskabc: I don't need the undefined value anymore, since we won't display the selector in such a case. I simplified it to just `this.props.indexedLanguages.length > 1` and refactored the `mapIndexedLanguages` function to exclude the undefined return value.

ledsoft (approved):
LGTM

PR #667, opened 2025-08-29
[Feat #664] Remember flat list selection in browser storage

### LINKED ISSUE #664: Remember flat list mode selection
As a user, I want TermIt to remember the flat list toggle value so that I don't have to trigger it every time I open term editing or any other view where a term selector is.

Implementation note: should use the same mechanism as remembering the UI language.

Haschman:


src/component/term/TermSelector.tsx
```diff
@@ -57,12 +58,13 @@ export const TermSelector: React.FC<{
   const terminalStates = useSelector(
     (state: TermItState) => state.terminalStates
   );
-  const [flatList, setFlatList] = React.useState(false);
   const treeSelect = React.useRef<IntelligentTreeSelect>(null);
-  const handleFlatListToggle = () => {
-    setFlatList(!flatList);
+
+  const flatList = useSelector((state: TermItState) => state.showTermsFlatList);
+  const handleFlatListToggle = React.useCallback(() => {
```
ledsoft: What is the point of memoizing this simple handler?

src/component/term/ParentTermSelector.tsx
```diff
@@ -187,9 +192,8 @@ export class ParentTermSelector extends React.Component<
   };
 
   private onFlatListToggle = () => {
-    this.setState({ flatList: !this.state.flatList }, () =>
-      this.treeComponent.current.resetOptions()
-    );
+    this.props.setTermsFlatList(!this.props.flatList);
+    this.treeComponent.current.resetOptions();
```
ledsoft: Won't this cause a double options reset? One here and the other in `componentDidMount` when the `flatList` prop value changes...

src/component/term/Terms.tsx
```diff
@@ -250,9 +251,8 @@ export class Terms extends React.Component<GlossaryTermsProps, TermsState> {
   };
 
   private onFlatListToggle = () => {
-    this.setState({ flatList: !this.state.flatList }, () =>
-      this.treeComponent.current.resetOptions()
-    );
+    this.props.setTermsFlatList(!this.props.flatList);
+    this.treeComponent.current.resetOptions();
```
ledsoft: Same question on double options reset as in `ParentTermSelector`.

ledsoft (commented):
A couple of inline comments asking for clarification.

ledsoft (approved):
LGTM

PR #671, opened 2025-09-18
[Feat #668] Put primary language first in languages tabs

### LINKED ISSUE #668: Put primary language first in languages tabs
When a term/vocabulary is multilingual, the primary language of the vocabulary should be first in the language tabs (currently, they are sorted alphabetically).

Haschman:


ledsoft (changes requested):
This does not do what the ticket requested - the selector should display first the **primary language of the vocabulary**, not the UI language.

Haschman:
> This does not do what the ticket requested - the selector should display first the **primary language of the vocabulary**, not the UI language.

Oh, alright, sorry. I was confused when I found only "requiredLanguage" and "language" in props, so the UI language made the most sense to me from a user perspective.

ledsoft (changes requested):
This is still not correct. It prioritizes the currently selected language, which in addition to not implementing the requested behavior leads to an issue when the tabs keep changing position when language is switched (see the attached video).

https://github.com/user-attachments/assets/c0d90982-beb3-4c49-b5b6-9e5b3861971a


A vocabulary has a `primaryLanguage` attribute which should be passed to the language selector (the `requiredLanguage` prop could be used for that or maybe rename it if the name seems confusing) and that language should be first in the list (i.e. be the first rendered tab). In the attached video, the vocabulary has `English` as primary language, so that should be the first tab.
If the issue description in the ticket was not clear enough, please ask for clarification before implementing something you are unsure is what was expected. It will save both of us time.

ledsoft (commented):
Two comments for consideration:
1. I think we should rename the `requiredLanguage` prop in `LanguageSelector`. In this case the languages cannot be removed, and I think the prop should be called `primaryLanguage` to better communicate its meaning.
2. `renderLanguages` and `renderRemovableLanguages` contain pretty much the same code, the only difference AFAIK being the rendering of the remove icon. Can this be refactored a little bit? The original way is not the best, but the new way seems unnecessarily wasteful. Maybe add an item rendering function that would be parameterized with boolean switch as to whether to render the remove icon or not. Both `renderLanguages` and `renderRemovableLanguages` would then call it for each item?

ledsoft (approved):
Excellent, thank you!

PR #676, opened 2025-09-26
[Feat #673] Added pagination to asset history

### LINKED ISSUE #673: Add paging to history
The history (`Change history` tab in asset detail) of some assets can get quite long. The current implementation just displays all the change records in one table which can get quite long. Most other parts of the application already support paging of the records (client-side, just a as a visual simplification).

As a user, I want change history to use paging to simplify navigation between change records and provide a more consistent UX.

Haschman:


src/component/changetracking/AssetHistory.tsx
```diff
@@ -93,10 +98,31 @@ export const AssetHistory: React.FC<AssetHistoryProps> = ({ asset }) => {
     filterAttribute,
     loadHistoryActionDebounced,
   ]);
+
+  React.useEffect(() => {
+    setPage(0);
+  }, [asset, filterAuthor, filterType, filterAttribute]);
+
+  React.useEffect(() => {
```
ledsoft: Under what circumstances is this effect necessary? Max page is controlled by `SimplePagination` so I don't see a reason to have a validation effect for page number (which is my understanding of this effect's functionality) and if we did need it, why not implement it as a wrapper for `setPage`?

Haschman: I wrote this for cases where, for example, you are on page 3 and then filter items, so that the list is reduced to fewer pages. However, I now realize that when any filter is changed, the page automatically resets to 0, which makes this effect unnecessary.

ledsoft (approved):
LGTM

PR #688, opened 2025-10-09
[Feat #678] Visual improvements of faceted search

### LINKED ISSUE #678: Visual improvements of faceted search
Currently, faceted search displays all the supported fields at once. Since adding support for custom attributes (#658), this list of fields can grow, making the UI a bit complicated (see the image).

Image

The goal is to rework the faceted search UI to make it more compact, displaying only the fields the user is interested it. An example could be a list of available properties, the user would check those they want to use, and the UI would display input fields only for these selected properties.

Haschman:


ledsoft (commented):
LGTM, just one minor layout issue I discovered - the input that has the "Exact match/Substring" switch does not align with input without this switch. Could you look into that before we merge this?

Image

ledsoft (approved):
LGTM

PR #699, opened 2025-10-27
[Enhancement kbss-cvut/termit#338] Introduce resource save reason for file uploads

### LINKED ISSUE #338: Make file backups more space-efficient
Since we backup Files on every changes, active use of the annotator tends to massively increase the number of these files. Especially when approving suggested occurrences creates new backups as well. This may cause problems for server backups because the Docker volumes TermIt uses are full of File backups.

We should consider making the file handling more space-efficient, for example by:
- reducing the number of backups, perhaps by making simple changes (like approving an occurrence) directly in place instead of making a copy of the file
- Compressing the backup files

lukaskabc:
Includes the file save reason when saving document file contents


ledsoft (approved):
LGTM

PR #706, opened 2025-11-02
add German translation

ctr49:


src/i18n/de.ts
```diff
@@ -0,0 +1,685 @@
...
+    "comments.comment.like.on": "Zustimmung entfernen",
+    "comments.comment.dislike": "Ich lehne diesen Kommentar ab",
+    "comments.comment.dislike.on": "Ablehnung entfernen",
+    "comments.comment.edited": "Bearbeitet",
+    "validation.message.tooltip": "Dies ist ein Ergebnis der Qualitätsüberprüfung. ",
+    "snapshots.title": "Snapshot-Verlauf",
+    "snapshots.created": "Snapshot erstellt",
+    "snapshots.show": "Snapshot anzeigen",
+    "snapshots.empty": "Keine vorherigen Snapshots gefunden.",
+    "snapshot.message": "Dieser {Typ} ist ein Snapshot und schreibgeschützt. ",
```
ledsoft: Please make sure that labels in `{}` are the same across languages, they are interpolated with actual values when displaying in the UI. In this particular case, it should be `{type}`.

ledsoft (changes requested):
Hi, thank you very much for wanting to contribute to TermIt. Could you please:
- Ensure labels in `{}` remain the same as in the other translation files (see the inline comment)
- Run Prettier (e.g. `npm run prettier:fix`) to ensure correct code formatting (it is normally supposed to run on commit, not sure why it did not in your case)

src/i18n/de.ts
ctr49: I see - 'search and replace' went a bit to far, apparently. I caught the unintentional changes and corrected them.

ledsoft (approved):
LGTM, thank you very much for providing the translation!

PR #714, opened 2025-11-17
[Feat #702] Extend custom attributes for annotating term relationships

### LINKED ISSUE #702: Extend custom attributes for annotating term relationships
Custom attributes need to be extended for the annotation of term relationships. This extension consists of:

- New domain of the custom attribute - "term relationships" (stored as `rdf:Statement`)
- For such custom attributes, the user can select from a set of properties representing term attributes to which the custom attribute will be applicable. The set of properties will consist of: 
  - basic term attributes, 
  - custom attributes whose domain and range are `skos:Concept`

Haschman:


src/component/administration/customization/CustomAttributeEdit.tsx
```diff
@@ -270,6 +388,50 @@ export const CustomAttributeEdit: React.FC = () => {
                 />
               </Col>
             </Row>
+            {domain === VocabularyUtils.RDF_STATEMENT && (
```
ledsoft, line 302: I suggest extracting the selector and its related code into a separate component, it basically doubled the amount of code in this file.

I also think it would make sense to allow adding new annotated relationships when editing an existing custom attribute, so it should not be disabled (just check when the user is editing that they are not removing any previously selected values).

ledsoft (commented):
See the inline comment for suggestions.

ledsoft (approved):
LGTM, the memos in AnnotatedRelationshipSelector seem like an overkill given there will typically be < 10 relationships to annotate but let's merge it as is.

PR #718, opened 2025-11-29
[Enhancement #679] Provide access to document backups

### LINKED ISSUE #679: Provide access to document backups
As a user, I want to be able to see backups of a file and be able to return to a selected version of the file. 

- The list should indicate the general type of the backup (text analysis invocation, added occurrence, periodic).
- When a backup is restored, TermIt should also generate corresponding term occurrences into the repository (much like when processing the results of text analysis).
- When the backup is restored, the system should first check whether the current version differs from the last available backup. If so, it should create a new backup before applying the selected one.

This follows kbss-cvut/termit#338.

lukaskabc:
closes #679 

- Adds Backups button to file list in vocabulary document
- The button shows modal with paged list of backups and button to restore them
- WRITE permission is required

- DE translation may need adjustments (automatic translation tool used)


src/component/asset/FileBackupList.tsx
```diff
@@ -0,0 +1,63 @@
...
+      <thead>
+        <tr>
+          <th>{i18n("date")}</th>
+          <th>{i18n("time")}</th>
+          <th>{i18n("resource.file.backup.reason")}</th>
+          <th></th>
+        </tr>
+      </thead>
+      <tbody>
+        {props.backups.map((backup) => {
```
ledsoft: You can get the index as the second argument to the `map` function.

src/i18n/en.ts
```diff
@@ -483,6 +487,17 @@ const en = {
     "resource.metadata.document.files.empty":
       "No files found. Start by adding some.",
     "resource.file.vocabulary.create": "Add File",
+    "resource.file.backup.reason": "Reason for creation",
```
ledsoft: "Backup reason"?

ledsoft:
One more thing I noticed when testing - it displays the 'Backups' button even for files that do not have content. This should not happen - only files that do have content should have the button that allows displaying the list of available backups.

lukaskabc:
What about hiding it when there are no backups available?

ledsoft:
> What about hiding it when there are no backups available?

Yeah, that probably makes sense too.

ledsoft (approved):
LGTM, resolved remaining comments myself.

PR #758, opened 2026-03-01
[Feat #713] Implement advanced search combining full-text search with faceted filtering

### LINKED ISSUE #713: Extend faceted search with FTS
# Popis
Faceted search should support full-text search as well, so that FTS results can be further filtered by facets.

# Využití
As a user, I want to use FTS but I want to filter the results using facets.

# Usefulness

2. It slows me down (I could run FTS and then go through the results manually, but this can be quite time consuming)
****

Haschman:
- faceted search replaced with new advanced search that combines full text search and faceted search


src/component/search/label/NavbarSearch.tsx
```diff
@@ -42,18 +42,16 @@ interface NavbarSearchState {
...
-const ROUTES_WITHOUT_SEARCH_OVERLAY = [
-  Routes.search,
-  Routes.searchTerms,
-  Routes.searchVocabularies,
-  Routes.facetedSearch,
-  Routes.publicSearch,
-  Routes.publicSearchTerms,
-  Routes.publicSearchVocabularies,
-];
+const ROUTES_WITHOUT_SEARCH_OVERLAY = [Routes.search, Routes.publicSearch];
```
ledsoft: This is the same as `ROUTES_WHERE_NAVBAR_SEARCH_HIDDEN`, isn't it? I don't expect these two to ever differ, so maybe we can use just one of the constants.

src/component/MainView.tsx
```diff
@@ -39,12 +42,22 @@ const VocabularyManagementRoute = React.lazy(
...
-);
-const SearchTerms = React.lazy(() => import("./search/SearchTerms"));
-const FacetedSearch = React.lazy(() => import("./search/facet/FacetedSearch"));
+const AdvancedSearch = React.lazy(() => import("./search/AdvancedSearch"));
+
+/**
+ * Registers a search listener so the navbar FTS overlay works on all pages.
+ * Replaces the old SearchTypeTabs behavior.
+ */
+function SearchListenerHelper() {
```
ledsoft: Looks like the same as in `public/MainView`. Why not reuse it (probably this one, or extract it into a separate file)?

src/component/search/AdvancedSearchResults.tsx
```diff
@@ -0,0 +1,87 @@
...
+import { Card, CardBody } from "reactstrap";
+import { useI18n } from "../hook/useI18n";
+import VocabularyUtils from "../../util/VocabularyUtils";
+import SearchResult from "../../model/search/SearchResult";
+import { SearchResultItem } from "./label/SearchResults";
+import TermResultItem from "./label/TermResultItem";
+import VocabularyResultItem from "./label/VocabularyResultItem";
+import SimplePagination from "../dashboard/widget/lastcommented/SimplePagination";
+
+export const RESULT_PAGE_SIZE = 20;
```
ledsoft: Maybe put it into `Constants` so that we have these constants in one place. Or better yet - reuse the page size configuration used for vocabularies list - it is stored in LocalStorage so we can consider it as the user's default paging setting...

ledsoft (commented):
A couple of minor suggestions, haven't tested the functionality yet.

ledsoft:
A couple more things I discovered when testing:
1. The search type toggle and input are bigger than usual throughout TermIt (they do no use the `small` bootstrap class). Is this intentional? I think it would be better if the UI consistently used the same size of buttons and inputs
image

2. When I search terms by other attributes than FTS, the results are shown with "Match found in attribute: {undefined}". I think it should not display this when no match field is available

3. When I search terms by type, the UI crashes with an error "Unsupported type of asset data". I think it is because searching by type causes the `skos:Concept` not to be present in the result items and the UI is thus unable to determine if the result is a term or a vocabulary.

Haschman:
About this:
> 
> 1. The search type toggle and input are bigger than usual throughout TermIt (they do no use the `small` bootstrap class). Is this intentional? I think it would be better if the UI consistently used the same size of buttons and inputs
> 
I agree with the toggle buttons and I'll fix those. However, the input field is larger because it’s the same size as the search input field in the nav bar. I like that they're a bit bigger than normal fields because they're more important and easier for the user to focus on. But if you want them smaller, we should probably scale down the one in the nav bar too.

ledsoft:
> About this:
> 
> > 1. The search type toggle and input are bigger than usual throughout TermIt (they do no use the `small` bootstrap class). Is this intentional? I think it would be better if the UI consistently used the same size of buttons and inputs
> 
> I agree with the toggle buttons and I'll fix those. However, the input field is larger because it’s the same size as the search input field in the nav bar. I like that they're a bit bigger than normal fields because they're more important and easier for the user to focus on. But if you want them smaller, we should probably scale down the one in the nav bar too.

Ok, let's make the buttons smaller and keep the input field large.

ledsoft (approved):
LGTM

PR #780, opened 2026-03-30
[Upd #770] Migrate react-table to v8

### LINKED ISSUE #770: Update react-table to tanstack/react-table
In connection with the planned implementation of the sheet-like vocabulary view (#751), we are considering using tanstack/react-table with react-query. However, currently TermIt uses v7 of react-table, so we need to migrate to the latest version.

Haschman:


ledsoft (approved):
LGTM

PR #786, opened 2026-04-14
[Feat #751] Sheet-like vocabulary display

### LINKED ISSUE #751: Sheet-like vocabulary display
As a user, I want to see the content of a vocabulary in a tabular manner so that I get an overview of all terms and their details in one place instead of having to navigate to individual term details.

Notes:
- The view should be paged, so that FCP does not take too long when the vocabulary contains a lot of terms
- I want to be able to hide columns so that the display is easier to navigate
- I want to be able to filter the terms in the view

Haschman:


src/component/vocabulary/VocabularySheetViewPage.tsx
```diff
@@ -0,0 +1,221 @@
...
+          >
+            <Button
+              id="sheet-view-create"
+              color="primary"
+              size="sm"
+              title={i18n("glossary.createTerm.tooltip")}
+              onClick={onCreateClick}
+            >
+              <GoPlus />
+              &nbsp;{i18n("glossary.new")}
```
ledsoft: I suggest using a margin on the icon instead of `nbsp`. We use a combination of both in the code base, but I am trying to shift to the margin.

src/component/vocabulary/VocabularySheetViewTable.tsx
```diff
@@ -0,0 +1,894 @@
...
+function resolveVocabularyFragment(vocabularyIri?: string): string {
+  if (!vocabularyIri) {
+    return "";
+  }
+  try {
+    return VocabularyUtils.create(vocabularyIri).fragment;
+  } catch {
+    return vocabularyIri;
+  }
+}
```
ledsoft: I suggest moving utility functions into a separate file.

src/component/vocabulary/VocabularySheetViewTable.tsx
```diff
@@ -0,0 +1,894 @@
...
+
+function resolveLocalizedTermLabel(
+  item: { iri?: string; label?: MultilingualString },
+  locale: string
+): string {
+  return (item.label ? getLocalized(item.label, locale) : "") || item.iri || "";
+}
+
+function previewTermList(
```
ledsoft: I suggest extracting the function/component into a separate file.

src/component/vocabulary/sheet/VocabularySheetViewPage.tsx
ledsoft, line 1: Maybe we should move the sheet-related components into a new directory

src/query/hook/useVocabularyTerms.ts
ledsoft, line 1: I suggest moving this outside the `component` directory - it is not a component. Maybe put it into he `query` directory, or create a dedicated one for react-query hooks?

ledsoft (commented):
A bunch of suggestions for now, going to test functionality. Also, you will have to sync package.json and package-lock.json, the build fails because they are out of sync...

ledsoft:
Additional feedback after testing it:
- Term type is not displayed correctly (see term detail for proper display)
- Allow expanding long texts (such as definition/scopeNote) - or maybe show them in a popup?
- Loading mask should be displayed when opening table view, right now the page just its open until the first batch of terms loads
- No need to show "Sheet view/Tabulkove zobrazeni" in page title - just the title of the vocabulary
- Term counter - replace "Loaded/Nacteno" with "Displayed/Zobrazeno" - it may be misleading when it does not display terms with missing label in the selected language - the term is loaded, it is just not displayed
- Subterms from a different vocabulary may not fit in the cell - only vocabulary badge is displayed
image

- When term label (first column) does not fit column width, the title should show the label. Now it shows "Open term detail" - maybe we should combine both
- Column order should correspond to attribute order in term detail - definition should be second after label

ledsoft (approved):
LGTM

PR #831, opened 2026-07-27
[Feat #752] Add altLables and hiddenLabels and optimize term updates

### LINKED ISSUE #752: Editing in the sheet-like vocabulary view
As a user, I want to be able to edit terms in the sheet-like vocabulary view. The UI should be by default in a read-only mode (so that I can, e.g., copy stuff from individual cells), but I want to be able to trigger editing of individual terms. Multiple changes to the same term should be possible before submit (auto-submit is probably not a good idea without undo), but editing multiple terms is not (it will be a part of another ticket).

Haschman:


ledsoft (commented):
I was testing the feature and in my case the sidebar would not disappear until terms are loaded from the server. The exact scenario:
1. I opened a vocabulary with approx. 50 terms
2. Edited related terms of a term
3. Hit save, sidebar is still visible while PUT request is being processed
4. Sidebar is still visible when PUT ended and GET for terms was sent
5. When data are received, sidebar is closed and I see the messages in the top-right corner

I think the sidebar should disappear once PUT finishes successfully, then the term loading runs in the background.

ledsoft (approved):
LGTM

PR #837, opened 2026-08-05
[Feat #753] Related vocabularies

### LINKED ISSUE #753: Related vocabularies
As a user, I want to be able to specify vocabularies related to another vocabulary (similar to imported vocabularies). The purpose of these is to restrict the selection of relatedMatch and exactMatch terms. However, I want to be able to override this behavior (so the behavior could be something like a pre-selected but editable filter).

Haschman:


src/i18n/cs.ts
```diff
@@ -400,6 +400,7 @@ const cs = {
     "vocabulary.detail.imports": "Importuje",
     "vocabulary.detail.imports.edit": "Importuje slovníky",
     "vocabulary.detail.document": "Dokument",
+    "vocabulary.detail.related": "Související slovníky",
```
ledsoft, line 403: I think we should introduce help text that explains the distinction between imported and related vocabularies and display it using the help icon similarly to how we do it for term attribute:

Image

The help text could be something like (for Czech):
```json
"vocabulary.detail.imports.help": "Importované slovníky slouží k tvorbě hierarchie mezi pojmy. Konkrétné, pojmy z importovaných slovníků mohou být použity jako nadřazené pojmy pojmů z tohoto slovníku.",
"vocabulary.detail.related.help": "Související slovníky slouží k předfiltrování výběru souvisejících pojmů a pojmů se stejným významem pouze na související slovníky. Toto předfiltrování lze vypnout."
```

src/component/public/vocabulary/VocabularySummary.tsx
```diff
@@ -22,6 +21,7 @@ import {
   getLocalizedOrDefault,
 } from "../../../model/MultilingualString";
 import { resolveInitialLanguage } from "../../vocabulary/VocabularySummary";
+import VocabulariesReferenceList from "../../../component/vocabulary/VocabulariesReferenceList";
```
ledsoft, line 24: Looks like an unnecessarily long import?

```suggestion
import VocabulariesReferenceList from "../../vocabulary/VocabulariesReferenceList";
```

src/component/term/TermSelector.tsx
```diff
@@ -109,11 +132,26 @@ export const TermSelector: React.FC<{
     <FormGroup id={id}>
       <div className="d-flex justify-content-between">
         {label}
+        {vocabularyIri && (
```
ledsoft, line 135: I wonder if we should also display a list of the related vocabularies, so that the user knows from which vocabularies the results will be?

ledsoft (commented):
See the inline comments, they are mostly suggestions.

ledsoft (approved):
LGTM

PR #845, opened 2026-08-16
[Feat #834] Add ability to adjust font size in tabular vocabulary view

### LINKED ISSUE #834: Font size adjustment in tabular vocabulary view
# Description
As a user, I want to see as much information as possible in the tabular vocabulary view. To do that, the font in the table should either be adjustable or should be smaller than it is now.

# Usage
Editing a vocabulary.

# Usefulness
2. The scenario can be realized without the proposed feature, but it is slow/cumbersome.
****

Haschman:


ledsoft (commented):
Two things:
1. The font size selector does not indicate current value on initial open when there is no value in the session storage. I.e. when I open the tabular view for the first time and open the font size selector, it does not indicate what the current font size is. Only after I select a value the selector shows it.
2. I would switch the columns and font size buttons - have the column selector first and font size second. Consider also moving the font size selector to the far right. While the other buttons and inputs are relevant to the data, this is just visualization control element and should probably be distinguished as such.

ledsoft (approved):
LGTM

PR #848, opened 2026-08-27
[Enhancement #820] Add support for custom attribute removal

### LINKED ISSUE #820: Delete custom attributes/smazání vlastních atributů
# Description

As a user, I want to be able to delete custom attributes. If the custom attribute is already used (i.e. it has values for some term), TermIt should notify me and ask whether the values should be removed as well, or I want to cancel the removal.

# Usage

Migration of custom attributes, their removal.

# Usefulness

1. The scenario cannot be realized without the proposed feature.
****

# Popis
Jako uživatel chci mít možnost odstranit vlastní atribut. Pokud již tento atribut je použit (tj. má u nějakého pojmu/slovníku hodnotu), musí mě na to TermIt upozornit a nechat mě rozhodnout, jestli mají být tyto hodnoty smazány, nebo chci akci zrušit.

lukaskabc:
#820

Backend PR: https://github.com/kbss-cvut/termit/pull/422

This PR adds remove buttons to the list of custom attributes, allowing them to be removed (administration > customization).

Before the attribute is removed, a confirmation modal is shown. 
If the attribute is already used, its usage is listed, and the user must confirm the removal by entering the attribute name in the input field.
The user can choose to keep the attribute usages and remove only the attribute itself.


src/action/AsyncCustomizationActions.ts
ledsoft, line 1: Maybe `AsyncCustomAttributeActions`? Or do you expect other kinds of customization actions?

src/action/AsyncCustomizationActions.ts
```diff
@@ -0,0 +1,56 @@
...
+
+export function loadCustomAttributeUsage(
+  attributeIri: IRI,
+  pageRequest: PageRequest
+) {
+  return (dispatch: ThunkDispatch) => {
+    const action = { type: ActionType.LOAD_CUSTOM_ATTRIBUTE_USAGE };
+    dispatch(asyncActionRequest(action, true));
+    return Ajax.getResponse(
+      Constants.API_PREFIX +
```
ledsoft:
```suggestion
      `${Constants.API_PREFIX}/data/custom-attributes/${attributeIri.fragment}/usage`
```

src/action/AsyncCustomizationActions.ts
```diff
@@ -0,0 +1,56 @@
...
+
+export function removeCustomAttribute(
+  attributeIri: IRI,
+  removeUsages: boolean
+) {
+  return (dispatch: ThunkDispatch) => {
+    const action = { type: ActionType.REMOVE_CUSTOM_ATTRIBUTE };
+    dispatch(asyncActionRequest(action, false));
+    return Ajax.delete(
+      Constants.API_PREFIX + `/data/custom-attributes/${attributeIri.fragment}`,
```
ledsoft:
```suggestion
      `${Constants.API_PREFIX}/data/custom-attributes/${attributeIri.fragment}`,
```

src/component/administration/customization/CustomAttributeRdfValueNode.tsx
ledsoft, line 1: Please remove all mentions of `RDF4J` from the PR. Stick to RDF(s) terminology, the fact that the backend internally uses RDF4J API should not leak into the frontend.

src/action/SyncActions.ts
```diff
@@ -216,3 +217,17 @@ export function setTermsFlatList(flatList: boolean): SetTermsFlatListAction {
     flatList,
   };
 }
+
+/**
+ * Tries to load saved page size from local storage.
+ * If no size is saved, falls back to {@link Constants.DEFAULT_PAGE_SIZE}
+ */
+export function getInitialPageSize() {
```
ledsoft: I am not sure it is a good idea to put this action here. `SyncActions` are expected to be invoked via Redux `dispatch`, but this function is not. I would prefer it to be moved into utilities (for example `UISettingsUtil`)

src/component/misc/table/Pagination.tsx
```diff
@@ -33,7 +34,9 @@ interface PaginationProps {
   allowSizeChange?: boolean;
 }
 
-const PAGE_SIZES = [10, 20, 30, 50];
+const PAGE_SIZES = [10, 20, 30, 50, Constants.DEFAULT_PAGE_SIZE].sort(
```
ledsoft: `Constants.DEFAULT_PAGE_SIZE` is not intended to be used for UI pagination, it is used for data fetching by the tree component (see the comment in `Constants`). Maybe we should rename the constant to `DEFAULT_FETCH_SIZE` then?

src/i18n/cs.ts
```diff
@@ -267,6 +268,12 @@ const cs = {
       "Vztahuje se na vlastnosti vztahů",
     "administration.customization.customAttributes.annotatedRelationships.help":
       "Vyberte, které vlastnosti vztahů mezi pojmy může tento vlastní atribut anotovat. Platí pouze když doména je 'Vztah mezi pojmy'.",
+    "administration.customization.customAttributes.removal.description":
+      'Opravdu chcete nenávratně odstranit atribut "{label}"?',
```
ledsoft:
```suggestion
      'Opravdu chcete odstranit atribut "{label}"?',
```

src/i18n/cs.ts
```diff
@@ -267,6 +268,12 @@ const cs = {
...
+    "administration.customization.customAttributes.removal.removeUsages":
+      "Odebrat všechny použití vlastního atributu.",
```
ledsoft:
```suggestion
      "Odebrat všechna použití vlastního atributu.",
```

src/i18n/cs.ts
```diff
@@ -267,6 +268,12 @@ const cs = {
...
+    "administration.customization.customAttributes.removal.confirm":
+      "Pro potvrzení odstranění vlastního atributu a všech jeho použití, zadejte název atributu",
```
ledsoft:
```suggestion
      "Pro potvrzení odstranění vlastního atributu a všech jeho použití zadejte název atributu",
```

ledsoft (changes requested):
Have not tested yet, just code review feedback.

src/action/AsyncCustomizationActions.ts
lukaskabc: The tab in the administration, as well as the directory for the related components, is named customization

ledsoft: Ok, in that case could you also move CustomAttribute-related async actions from `AsyncActions` into this file so that they are in one place?

src/component/misc/table/Pagination.tsx
lukaskabc:
```ts
// Size of page fetched from server
DEFAULT_PAGE_SIZE: 100,
```
The comment still matched my usecase.
Currently the value is being used for tree select components but also for [default page size in advanced search](https://github.com/kbss-cvut/termit-ui/blob/cf971c87415e4d2135dde7e806a41774df0176cb/src/action/SearchActions.ts#L244)

So should I rename it to `DEFAULT_TREE_SELECT_FETCH_SIZE` and introduce a different constant for the default page size?

ledsoft: But in the advanced search it is again fetch size, not size in the page size selector. I agree with renaming the constant. I am not sure it is necessary to introduce a new default page size, I would expect it to be one of the existing page size options.

ledsoft (approved):
LGTM

PR #856, opened 2026-09-02
[Feat #754] Add ability to batch edit terms

### LINKED ISSUE #754: Batch term editing
As a user, I want to be able to make batch changes to the terms in a vocabulary. The scenario is that I am in the sheet-like view of a vocabulary and I want to add a particular related term to multiple terms in the vocabulary. Let's say I have multiple domain domain terms that are to be connected to terms representing their attributes (connected via skos:relatedMatch). In this case I may want to select multiple terms and add an attribute to all of them.

Required functionality:
- Add relatedMatch, exactMatch, broader, type value

Haschman:


src/query/hook/useBatchEditTerms.ts
```diff
@@ -0,0 +1,145 @@
...
+import Ajax, { content } from "../../util/Ajax";
+import VocabularyUtils from "../../util/VocabularyUtils";
+import { queryKeys } from "../queryKeys";
+import Term from "../../model/Term";
+import { publishMessage } from "../../action/SyncActions";
+import { createFormattedMessage } from "../../model/Message";
+import MessageType from "../../model/MessageType";
+
+interface BatchEditParams {
+  apiPrefix: string;
```
ledsoft: API prefix variable is not necessary here. It is used in the application only in parts where both public (read-only) and authenticated APIs can be accessed (like term/vocabulary reading), but for editing the prefix is always the same - the authenticated API.

src/component/vocabulary/sheet/sidebar/SelectedTermsList.tsx
```diff
@@ -0,0 +1,151 @@
...
+                className="d-flex justify-content-between align-items-center"
+                onClick={isExpandable ? toggleExpand : undefined}
+                style={{ cursor: isExpandable ? "pointer" : "default" }}
+              >
+                <span className="mb-0">
+                  {getLocalizedInLanguage(term.label, language) || term.iri}
+                </span>
+                {isExpandable && (
+                  <span className="text-muted">
+                    {isExpanded ? <FaMinus size={14} /> : <FaPlus size={14} />}
```
ledsoft, line 138: Could you add a tooltip indicating what the button means (especially the + could be confusing). A regular HTML `title` would suffice.

ledsoft (commented):
One issue I discovered:
- When _Limit to related_ is enabled (which it is by default) in the exact match selector, I disable it, select a term, then switch to editing another property, then switch back to exact matches, the selected value is gone because the _Limit to related_ is again enabled and the selected term is not loaded.
One solution I can think of is if a value is passed to the exact match selector (or maybe any term selector for that matter) and it does not match any of the related vocabularies, the _Limit to related_ should be disabled so that the value can be loaded.

Haschman:
> * When _Limit to related_ is enabled (which it is by default) in the exact match selector, I disable it, select a term, then switch to editing another property, then switch back to exact matches, the selected value is gone because the _Limit to related_ is again enabled and the selected term is not loaded.

Ah, I did not realise that the selected value is kept when switching the selected property. I can fix it like you wrote but I think a better way is just always clear the selected value on switch.

ledsoft (approved):
LGTM, I fixed incorrect API constant reference.

